PR #55, opened 2015-02-18
Add `passthrough` option, fixes #25, fixes gulpjs/gulp#840

### LINKED ISSUE #25: Passing gulp.src() to gulp.pipe() is erroring
My gulp task:

```
gulp.task('build', ['sass'], function (done) {
	var child = spawn('./node_modules/.bin/r.js', ['-o', 'app/assets/js/app.build.js'], {
		stdio: 'inherit'
	});

	child.on('close', function () {
		gulp.src('app/assets/build/js/main.js')
			.pipe(gulp.dest('app/assets/build'))
			.pipe(gulp.src('app/assets/build/js', {read: false}))
			.pipe(plugins.clean())
			.on('end', done);
	});
});
```

My error:

```
stream.js:94
      throw er; // Unhandled stream error in pipe.
            ^
Error: write after end
    at writeAfterEnd (/Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:145:12)
    at Transform.Writable.write (/Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:193:5)
    at Stream.ondata (stream.js:51:26)
    at Stream.EventEmitter.emit (events.js:95:17)
    at queueData (/Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/node_modules/map-stream/index.js:43:21)
    at next (/Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/node_modules/map-stream/index.js:71:7)
    at /Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/node_modules/map-stream/index.js:85:7
    at done (/Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/lib/dest/writeContents/index.js:9:5)
    at /Users/callumacrae/Sites/finance-finder/node_modules/gulp/node_modules/vinyl-fs/node_modules/graceful-fs/graceful-fs.js:102:5
    at Object.oncomplete (evalmachine.<anonymous>:107:15)

~/Sites/finance-finder rc-2*  14s
❯ gulp --version
[17:57:19] CLI version 3.8.0
[17:57:19] Local version 3.8.0
```

UltCombo:
Ref: gulpjs/gulp#840 #25

### DIFF
--- a/lib/src/index.js
+++ b/lib/src/index.js
@@ -20,13 +20,14 @@ function src(glob, opt) {
     read: true,
     buffer: true
   }, opt);
+  var pass, inputPass;
 
   if (!isValidGlob(glob)) {
     throw new Error('Invalid glob argument: ' + glob);
   }
   // return dead stream if empty array
   if (Array.isArray(glob) && glob.length === 0) {
-    var pass = through.obj();
+    pass = through.obj();
     if (!options.passthrough) {
       process.nextTick(pass.end.bind(pass));
     }
@@ -50,7 +51,7 @@ function src(glob, opt) {
   }
 
   if (options.passthrough) {
-    var inputPass = through.obj();
+    inputPass = through.obj();
     outputStream = duplexify.obj(inputPass, merge(outputStream, inputPass));
   }
 
--- a/lib/src/index.js
+++ b/lib/src/index.js
@@ -4,6 +4,8 @@ var assign = require('object-assign');
 var through = require('through2');
 var gs = require('glob-stream');
 var File = require('vinyl');
+var duplexify = require('duplexify');
+var merge = require('merge-stream');
 
 var filterSince = require('./filterSince');
 var getContents = require('./getContents');
@@ -18,20 +20,21 @@ function src(glob, opt) {
     read: true,
     buffer: true
   }, opt);
-  var pass = through.obj();
 
   if (!isValidGlob(glob)) {
     throw new Error('Invalid glob argument: ' + glob);
   }
   // return dead stream if empty array
   if (Array.isArray(glob) && glob.length === 0) {
-    process.nextTick(pass.end.bind(pass));
+    var pass = through.obj();
+    if (!options.passthrough) {
+      process.nextTick(pass.end.bind(pass));
+    }
     return pass;
   }
 
   var globStream = gs.create(glob, options);
 
-  // when people write to use just pass it through
   var outputStream = globStream
     .pipe(through.obj(createFile))
     .pipe(getStats(options));
@@ -46,7 +49,12 @@ function src(glob, opt) {
       .pipe(getContents(options));
   }
 
-  return outputStream.pipe(pass);
+  if (options.passthrough) {
+    var inputPass = through.obj();
+    outputStream = duplexify.obj(inputPass, merge(outputStream, inputPass));
+  }
+
+  return outputStream;
 }
 
 function isValidGlob(glob) {
--- a/README.md
+++ b/README.md
@@ -59,9 +59,9 @@ fs.src(['*.js', '!b*.js'])
     - Default value is `true`
     - `false` will disable writing the file to disk via `.dest()`
   - since - `Date` or `number` if you only want files that have been modified since the time specified.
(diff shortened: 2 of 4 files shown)


lib/src/index.js
```diff
@@ -46,7 +49,12 @@ function src(glob, opt) {
       .pipe(getContents(options));
   }
 
-  return outputStream.pipe(pass);
+  if (options.passthrough) {
+    var inputPass = through.obj();
```
contra: var in an if condition should fail the linter

contra:
How does the ordering work? Does it wait for the parent to finish, then emit? Or is it just first come first serve

lib/src/index.js
UltCombo: What linter? The project doesn't have JSCS, and JSHint has moved away from checking code style I believe. Anyway, should I move the `var` declarations to the top of their scope?

UltCombo:
@contra the PR uses merge-stream, so it is first come first served -- this is [gulp-add-src's default](https://github.com/urish/gulp-add-src/blob/87bfcca62725a6bc19185f8e1bf60526277f546b/index.js#L19-L22). This is also why I'm using [`.some` in some tests](https://github.com/UltCombo/vinyl-fs/commit/b9629ea6e9b9b797b518ab8192f4a2ec446ea325#diff-5a25c8b6edf4c2b15264b5b2a8414ec4R299).

gulp-add-src also provides methods for prepend/append, but I don't know if those are worth adding to vinyl-fs -- users that need ordered streaming can still use gulp-add-src I believe.

I'm also a bit concerned that there is no deduping in their (and, consequently, our) stream merging logic. This is most likely a rare enough use case that can be addressed in the future in case such use cases actually surface.

contra:
@UltCombo I think we should support some form of ordering and deduplicaton

UltCombo:
@contra which form of ordering should we prefer (or should ordering be customizable)? Not sure about deduping, it might be over-thinking it a bit too much -- I believe it is mostly a corner case, and can be cleanly solved in user code with a simple `.pipe(require('unique-stream')('path'))` once we implement ordering.

UltCombo:
The main problem I see with supporting ordering and deduping options is that it'd be too much API pollution for unnecessary/simple abstractions.

Might as well recommend people to use gulp-add-src instead of implementing `passthrough: true` (plus unique-stream when they want), dunno.

lib/src/index.js
```diff
@@ -18,20 +20,21 @@ function src(glob, opt) {
...
   }, opt);
-  var pass = through.obj();
 
   if (!isValidGlob(glob)) {
     throw new Error('Invalid glob argument: ' + glob);
   }
   // return dead stream if empty array
   if (Array.isArray(glob) && glob.length === 0) {
-    process.nextTick(pass.end.bind(pass));
+    var pass = through.obj();
```
contra: var in an if block

lib/src/index.js
contra: @UltCombo Yeah I think you can merge inputPass and pass into the same var

lib/src/index.js
UltCombo: @contra I don't really know what code style you want.

UltCombo:
@contra I don't really see what is wrong with `var` inside block, it is perfectly valid JS and if anything, as long as the given binding is only referenced inside the given block, eases the transition to ECMAScript 2015's block bindings (`s/var/let/g` or `s/var/const/g`).

Anyway, I've moved the `var` declarations to outside of blocks as requested.

contra:
@UltCombo It makes it harder to follow things imo - having vars at the top makes it easier to know what is in scope

UltCombo:
Alright, no prob. `=]`

contra:
Available on the gulp 4.0 branch now

PR #57, opened 2015-02-20
Follow symlinks

### LINKED ISSUE #39: Symlinked directories cause EISDIR error from `vinyl-fs.src()`
The fix for #38 is causing me issues when I have symlinked directories. In my use case I want to follow symlinks and copy their contents, etc. However the use of `lstat` causes `vinyl-fs` to think symlinked directories are files and I think it tries to load their content resulting in EISDIR errors.

I have created the following project to demonstrate the problem

https://github.com/pghalliday/test-vinyl-fs

valeriangalliat:
* Removed `getStats` function.
* Added a `resolveSymlinks` function that exposes a `stat` property on file object as a side effect (making `getStats` obsolete).
* Added tests.
* Fixes #39.


contra:
Should this be the default behavior or behind a flag?

Also code coverage decreased from 100%

valeriangalliat:
> Should this be the default behavior or behind a flag?

I'd prefer to leave this decision to someone that knows vinyl-fs better than me. :smile:

> Also code coverage decreased from 100%

Before:

```js
if (stat) {
  file.stat = stat;
}
cb(err, file);
```

After:

```js
if (err) { // not covered
  return cb(err);
}
file.stat = stat;
cb(null, file);
```

That's roughly the kind of changes I had to make when switching from `getStats` to `resolveSymlinks`, because while the first format allows to increase code coverage without having to explicitely test the error case, it makes the code less readable, mostly when you have nested calls like this.

I have even no idea what I have to pass to `vfs.src` to make `fs.lstat` and `path.realpath` "return" an error… if I pass an unexisting file, it will be handled at the glob level and will never reach the `resolveSymlinks` function. Any idea?

contra:
Hmm not really sure how to test those. I guess you would have to mock the fs module and have it return errors and make sure they bubble up. Seems like overkill though IMO - thanks for the PR

contra:
Will be published as a part of the 4.0 release when it is ready. 4.0 = master branch right now

PR #59, opened 2015-02-23
Implemented `overwrite` option in `dest()`

derekslife:
This is in reference to: https://github.com/wearefractal/vinyl-fs/issues/30

It seemed like the overwrite option was stubbed out, but the `wx` flag was never utilized in the fs operations. This commit wires it up, handles the error catch scenario and adds some tests. Let me know what you think!

### DIFF
--- a/lib/dest/writeContents/index.js
+++ b/lib/dest/writeContents/index.js
@@ -31,7 +31,9 @@ function writeContents(writePath, file, cb) {
   }
 
   function written(err) {
-    if (err) {
+    var ignorableError = (err && err.code === 'EEXIST' && file.flag === 'wx');
+
+    if (err && !ignorableError) {
       return complete(err);
     }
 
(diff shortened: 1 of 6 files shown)


lib/dest/writeContents/index.js
```diff
@@ -31,7 +31,9 @@ function writeContents(writePath, file, cb) {
   }
 
   function written(err) {
-    if (err) {
+    var ignorableError = (err && err.code === 'EEXIST' && file.flag === 'wx');
```
contra: canIgnoreError or isErrorFatal would be more descriptive imo

contra:
Very clean - one minor nitpick and I am +1 on merging this. Thanks!

derekslife:
Thanks!

contra:
Available on master - will be published in the 4.0 release

jlukic:
:+1:

Narretz:
Does "will be published in the 4.0 release" mean in gulp 4.0? Because that would mean this feature is documented since over a year ago, but not available yet ...

phated:
@Narretz comments like that are not welcome here. Feel free to use the module directly.

PR #61, opened 2015-03-05
Compare modes on comparable bases

Meroje:
Converting only the source mode will yield a diiference that results in always trying to chmod the destination, which may be forbidden in some cases (in my case the conversion resulted in a missing flag which I wasn't allowed to remove).

### DIFF
--- a/lib/dest/writeContents/index.js
+++ b/lib/dest/writeContents/index.js
@@ -44,9 +44,7 @@ function writeContents(writePath, file, cb) {
       if (err) {
         return complete(err);
       }
-      // octal 7777 = decimal 4095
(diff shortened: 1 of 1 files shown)


contra:
CI failed

Meroje:
Ok, now to see why it's passing on the work machine.

Meroje:
screw this, it won't answer the problem.
So here is my issue : stat on the terminal gives a mode of 2775 on a folder, but file.stat.mode is 42775 and this extra flag is messing the comparison on octal 0777 here https://github.com/wearefractal/vinyl-fs/blob/db025fe85b190414b28512c1f5ef08662564d4eb/lib/dest/writeContents/index.js#L48.

PR #62, opened 2015-03-06
Fix mode comparison

Meroje:
Related to #61, I found that the actual issue is currentMode gets its flags stripped by the bitwise comparison, but not the destination mode. This adds bitwise comparison to the dest mode to fix detection of different modes.

### DIFF
--- a/lib/dest/writeContents/index.js
+++ b/lib/dest/writeContents/index.js
@@ -44,12 +44,12 @@ function writeContents(writePath, file, cb) {
       if (err) {
         return complete(err);
       }
-      // octal 7777 = decimal 4095
-      var currentMode = (st.mode & 4095);
-      if (currentMode === file.stat.mode) {
+      var currentMode = (st.mode & parseInt('0777', 8));
+      var expectedMode = (file.stat.mode & parseInt('0777', 8));
+      if (currentMode === expectedMode) {
         return complete();
       }
-      fs.chmod(writePath, file.stat.mode, complete);
+      fs.chmod(writePath, expectedMode, complete);
     });
   }
 
--- a/test/dest.js
+++ b/test/dest.js
@@ -713,6 +713,56 @@ describe('dest stream', function() {
     stream.end();
   });
 
+  it('should see a file with special chmod (setuid/setgid/sticky) as matching', function(done) {
+    var inputPath = path.join(__dirname, './fixtures/test.coffee');
+    var inputBase = path.join(__dirname, './fixtures/');
+    var expectedPath = path.join(__dirname, './out-fixtures/test.coffee');
+    var expectedContents = fs.readFileSync(inputPath);
+    var expectedCwd = __dirname;
+    var expectedBase = path.join(__dirname, './out-fixtures');
+    var expectedMode = 03722;
+    var normalMode = 0722;
+
+    var expectedFile = new File({
+      base: inputBase,
+      cwd: __dirname,
+      path: inputPath,
+      contents: expectedContents,
+      stat: {
+        mode: normalMode
+      }
+    });
+
+    var expectedCount = 0;
+    spies.setError(function(mod, fn) {
+      if (fn === 'stat' && arguments[2] === expectedPath) {
+        expectedCount++;
+      }
+    });
+
+    var onEnd = function(){
+      expectedCount.should.equal(1);
+      should(chmodSpy.called).be.not.ok;
+      realMode(fs.lstatSync(expectedPath).mode).should.equal(expectedMode);
+      done();
+    };
+
+    fs.mkdirSync(expectedBase);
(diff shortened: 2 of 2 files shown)


contra:
got a test?

Meroje:
Do you mean a test to prove current implementation breaks, or that it works in the PR ?

contra:
@Meroje - both, a test that fails currently and passes after this PR

Meroje:
I've got a failing test [here](https://travis-ci.org/Meroje/vinyl-fs/builds/53669131), passing on this pull request.

lib/dest/writeContents/index.js
```diff
@@ -44,12 +44,12 @@ function writeContents(writePath, file, cb) {
       if (err) {
         return complete(err);
       }
-      // octal 7777 = decimal 4095
-      var currentMode = (st.mode & 4095);
-      if (currentMode === file.stat.mode) {
+      var currentMode = (st.mode & parseInt('0777', 8));
+      var expectedMode = (file.stat.mode & parseInt('0777', 8));
+      if (currentMode === expectedMode) {
```
contra: hmm this part doesnt click with me. if the mode is the same then the file contents wont get updated?

Meroje: parseInt is just there for readability, parseInt('07777', 8) would give 4095.  
I changed it to 0777 to remove special bits, for example : 
- given a folder with sticky bit and setuid (5722)
- if you then make a file with mode 644, it will become 5644
- as of master, `currentMode === file.stat.mode` evaluates to false
- vinyl-fs then tries to chmod the file, which fails when you are not the owner

My opinion is to let the filesystem handle the special bits and just deal with the actual permissions

contra: @Meroje That wasn't my concern at all, not sure where that came from. My concern is that this is skipping file writes if the mode is the same, instead it should just not pass a mode into the file write if the mode is the same

Meroje: The file is written before the chmod check, written() is the callback of the write* https://github.com/wearefractal/vinyl-fs/blob/ba57ce76e978850f79f78bba9b8aa1a73ac5fb2f/lib/dest/writeContents/index.js#L10-22

contra: Ah sorry I misread the diff - saw line 52 as fs.writeFile

contra:
This is causing a failing test for me locally

```
  1) dest stream should see a file with special chmod (setuid/setgid/sticky) as matching:

      Uncaught AssertionError: expected 978 to be 2002
      + expected - actual

      +2002
      -978

      at Assertion.fail (/Users/contra/Projects/vinyl-fs/node_modules/should/lib/assertion.js:196:17)
      at Assertion.prop.(anonymous function) (/Users/contra/Projects/vinyl-fs/node_modules/should/lib/assertion.js:81:17)
      at DestroyableTransform.onEnd [as _flush] (/Users/contra/Projects/vinyl-fs/test/dest.js:746:56)
      at DestroyableTransform.<anonymous> (/Users/contra/Projects/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:135:12)
      at DestroyableTransform.g (events.js:257:16)
      at emitNone (events.js:67:13)
      at DestroyableTransform.emit (events.js:163:7)
      at finishMaybe (/Users/contra/Projects/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:371:12)
      at endWritable (/Users/contra/Projects/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:378:3)
      at DestroyableTransform.Writable.end (/Users/contra/Projects/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:356:5)
      at DestroyableTransform.onend (stream.js:60:10)
      at emitNone (events.js:72:20)
      at DestroyableTransform.emit (events.js:163:7)
      at /Users/contra/Projects/vinyl-fs/node_modules/through2/node_modules/readable-stream/lib/_stream_readable.js:965:16
      at process._tickCallback (node.js:349:13)
```

Meroje:
Just saw it, the the chmod changes to 01722 on [stream.end](https://github.com/wearefractal/vinyl-fs/blob/fd9388db1a15afe1e3c748d3859ab969c34ba1db/test/dest.js#L763).
The whole thing being to not care about the first bit, I would change the expectedMode or remove the line `realMode(fs.lstatSync(expectedPath).mode).should.equal(expectedMode);` (was copy/pasted)

contra:
@Meroje PR?

Meroje:
It seems this is an osx thing, the test still passes on my ubuntu work machine, I will just remove the chmod equal expectation.

PR #63, opened 2015-03-19
export `filterSince` to `vinyl-filter-since`

tunnckoCore:
```
1) source stream should pass through writes:
     Uncaught Error: File not found with singular glob
```
fails before the changes :) look at it.

I tried to resolve it, but I'm no so deep here.

Extract it because it can be useful. Sorry if already have module for that, but I didn't find.


contra:
Looks fine but https://github.com/tunnckoCore/vinyl-filter-since/blob/master/index.js#L17 doesn't make any sense

tunnckoCore:
It is for debuging and testing, actually higher coverage. Any other/better ideas?
Hmm. Is there some env that travis enables and tell us it running on travis machine? huh, i don't know.

contra:
@tunnckoCore How does that improve test coverage?

tunnckoCore:
@contra I mean test coverage of the `vinyl-filter-since` pkg, not vinyl-fs.

tunnckoCore:
ping @contra

contra:
Still not digging that debug option

tunnckoCore:
@contra okey if I remove it? coverage of `vinyl-filter-since` package will be lower...

contra:
@tunnckoCore I don't understand how it is helping coverage by having that in there

tunnckoCore:
look/run the tests? this one 
```
should not work properly when time is in future
```
this possibility cant be handled/tested if there's no some trick like `debug` "option"

tunnckoCore:
or.... maybe if we add something to the `file` object in this `if` [here](https://github.com/tunnckoCore/vinyl-filter-since/blob/master/index.js#L14-L16)
then check for it.. hm.

tunnckoCore:
no no, cant.

tunnckoCore:
got it? by default it won't go to the next pipe if given date is in future, so this "debug option" help us to determine that scenario and bump coverage.

so the coverage of `vinyl-filter-since` package would be
```
Statements   : 50% ( 17/34 )
Branches     : 25% ( 1/4 )
Functions    : 38.46% ( 5/13 )
Lines        : 50% ( 17/34 )

```

doowb:
@tunnckoCore take a look at this test https://github.com/wearefractal/vinyl-fs/blob/master/test/src.js#L251-L265

The filter since pieces have 100% code coverage here so it shouldn't be hard to do the same in `vinyl-filter-since`

tunnckoCore:
actually it's not correct and [the given date here](https://github.com/wearefractal/vinyl-fs/blob/master/test/src.js#L253) is not in the future.

tunnckoCore:
another way without debug option is to pass ` cb(null, false);` after the if, but it will break next pipe/plugin

but debug option is more elegant and more correct to be done

doowb:
All you have to do is make sure the file didn't go through the pipe. And the current test does that.

jonschlinkert:
just curious. if this is externalized, will there be a reason for other modules to use `vinyl-filter-since`, or will it only be used by vinyl-fs?

tunnckoCore:
> will there be a reason for other modules

maybe, why not?

jonschlinkert:
ah, sorry I see that it's just a plugin. I have no clue what I was looking at earlier

tunnckoCore:
okey, its ready, I'll publish in a bit, and will update this PR.

@jonschlinkert no prob.

tunnckoCore:
done

tunnckoCore:
@contra refactored and no more `debug` option + 100% coverage :)
https://github.com/tunnckoCore/vinyl-filter-since

contra:
Test failng? ^

tunnckoCore:
lol, it fails before that PR, check travis. Even few days before my first pr

tunnckoCore:
cool, thanks :)

PR #64, opened 2015-03-22
switch to glob-watcher 2.0

phated:
this relies on the recent breaking changes to glob-watcher being published to npm as 2.0

@contra this will cause vinyl-fs to require a 2.0 bump also.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "duplexify": "^3.2.0",
     "glob-stream": "^4.0.1",
-    "glob-watcher": "^0.0.8",
+    "glob-watcher": "^2.0.0",
     "graceful-fs": "^3.0.0",
     "merge-stream": "^0.1.7",
     "mkdirp": "^0.5.0",


contra:
Will bump vinyl-fs to 2 after I finish some other fixes

phated:
Sounds good to me :birthday:

timelf123:
I'm hitting an issue where vinyl-fs' version of gaze@0.5.x has an issue when setting `maxListeners: 0`. It should set them to infinite but currently ignores the option (but fixed on 0.6.x and higher). I am passing in `maxListeners` as a workaround:

`gulp.watch(files, { maxListeners: 999 }, ['taskName'])`

contra:
@timelf123 You should open an issue on gaze and ask them to push a fix to the 0.5 branch as a patch release then.

timelf123:
Done: https://github.com/shama/gaze/issues/193

timelf123:
Is it possible to bump the version of gaze required since the issue has been fixed on 0.6? Seems silly to backport a fix when the next minor version took care of it

contra:
@timelf123 the 0.6 branch is full of bugs and doesn't work, this is all over the gaze project and the gulp project - 0.6 is broken, nobody can use it. It may have fixed some bugs but it introduced more issues than it solved.

timelf123:
My mistake, I didn't even test :P Just read that that specific bug was fixed. Thanks for clarifying

Issue #23, opened 2014-06-10
encoding param


contra:
should work on src and dest, default being utf8

czbo:
I am currently trying to use gulp.src to read and transform html files with Chinese (big5) characters into utf8 encoding. Doesn't seem completely straight forward; still trying to figure it out...

czbo:
Follow up to my own comment yesterday: I am able to figure out how to read in a file with big5 encoding and output to utf8 encoding. It doesn't work if I simply specify {encoding: 'utf8} in the .dest function; I need to use convertEncoding to process the Stream and then do other things. It seems to work the same even if the option {encoding:'big5'} is omitted in the .src function.

<other required plugins omitted...>
var convertEncoding = require('gulp-convert-encoding');
gulp.task('encode', function () {
  return gulp.src('app/**/*.html',{encoding: 'big5'})
    .pipe(convertEncoding({from: 'big5', to: 'utf8'}))
    .pipe(gulp.dest('.tmp'))
});

phated:
@contra do you have any thoughts on how to implement this?

contra:
Haven't looked into it yet - one thing to keep in mind is concating files of multiple encodings together. I've been wrangling with some issues in gulp-concat about that.

phated:
maybe vinyl should have an encoding property that plugins like gulp-concat could check and emit an error if they see different encodings?

phated:
@contra any thoughts on my comment above?

alvieirajr:
https://www.npmjs.com/package/gulp-convert-encoding

phated:
What does this API look like?  Was thinking about implementing with `iconv-lite` which would give us a stripBOM replacement too but I don't have a good sense of how this works.

phated:
@contra I'm thinking we should consider shipping the encoding support that Erik already has completed in #287

contra:
@phated I agree - the PR looks good to me and seems like a minor release which is nice. Is the only outstanding item to propagate the new option docs to the main docs?

phated:
I'd be more comfortable releasing it in the major to avoid any weird edge cases we haven't considered.

PR #66, opened 2015-04-27
Expose readFile encoding option through vfs.src

limdauto:
Here is a PR for `src` regarding #23.

I will add support for `dest` if this is up to par :)


lib/src/getContents/index.js
```diff
@@ -14,7 +15,9 @@ function getContents(opt) {
 
     // read and pass full contents
     if (opt.buffer !== false) {
-      return bufferFile(file, cb);
+      // set encoding type if user specifies it
+      var options = 'encoding' in opt ? {'encoding': opt.encoding} : null;
```
callumacrae: Why not just `var options = { encoding: opt.encoding }`? If `opt.encoding` is undefined, it'll be passed through as undefined, which is fine.

callumacrae:
lgtm except for the one thing

lib/src/getContents/index.js
limdauto: Thanks. Fixed.

contra:
Test fail?

limdauto:
@contra it's failing on master

lib/src/getContents/bufferFile.js
```diff
@@ -3,12 +3,18 @@
...
-  fs.readFile(file.path, function(err, data) {
+function bufferFile(file, opt, cb) {
+  fs.readFile(file.path, opt, function(err, data) {
     if (err) {
       return cb(err);
     }
-    file.contents = stripBom(data);
+    // if user specifies encoding type, a string is returned
+    // so we need to convert it back to buffer
+    if (!Buffer.isBuffer(data) && typeof data === 'string') {
```
contra: just typeof is sufficient here, no need for both checks. then you can make this a ternary

lib/src/getContents/index.js
```diff
@@ -14,7 +15,9 @@ function getContents(opt) {
 
     // read and pass full contents
     if (opt.buffer !== false) {
-      return bufferFile(file, cb);
+      // set encoding type if user specifies it
+      var options = {'encoding': opt.encoding};
```
contra: objects are multi-line, also encoding doesnt need quotes around it

lib/src/getContents/index.js
```diff
@@ -14,7 +15,9 @@ function getContents(opt) {
...
+      return bufferFile(file, options, cb);
```
contra: why not just pass the whole options object through?

contra:
How does this work with writing files out? Aren't we going to need to specify the encoding there as well

lib/src/getContents/index.js
limdauto: I was trying to be explicit with the options so it matches readFile API.

lib/src/getContents/bufferFile.js
limdauto: That's true

limdauto:
@contra I understand without passing encoding to writing out, all this is kind of pointless. I just made the PR to see if I'm on the right track. Will add `dest` shortly.

lib/src/getContents/bufferFile.js
contra:
```js
if (typeof data === 'string) {
  data = new Buffer(data, opt.encoding);
}
file.contents = stripBom(data);
```

would be cleanest imo

lib/src/getContents/index.js
contra: @limdauto passing the whole thing down is what is done for everything, makes it easier to add new option behavior later on down the road

sindresorhus:
What use-case is this solving exactly? When would you ever need to be able to specify `ascii` encoding?

phated:
Encoding doesn't really make sense because we always read in as buffer, which doesn't have an encoding.  Plugins use `toString()` on that buffer to turn it into a string, so there's not a good way to expose the encoding information without breaking all plugins.

sindresorhus:
I would think https://github.com/heldinz/gulp-convert-encoding would be a better solution if you're using obscure encodings.

limdauto:
Hi guys, so sorry I completely forgot about this PR. It was more than a month ago and I have indeed found out that it's better to convert my stuff to sane encoding first rather than requiring it from gulp/vinyl-fs. Thanks for your comments. I'm closing this.

PR #71, opened 2015-06-11
Update vinyl to 0.5.x

lijunle:


lijunle:
I might think it could be better to pop up vinyl-fs version if merged. :)

contra:
@lijunle vinyl-fs has some other stuff to be finished on master before the next bump and release

lijunle:
@contra after release the new version, it could be great to leverage the new one in gulp.

PR #73, opened 2015-07-01
using is-valid-glob module to handle nested arrays properly

doowb:
If a nested array: `[ ['./fixtures/*.coffee'] ]` is passed, the current `isValidGlob` method will return `true` and `glob-stream` will end up throwing an 'Invalid glob at index 0` error.

Using `is-valid-glob` handles this case and causes `src` to throw an `Invalid glob argument: ` error instead.

Added test to validate.


contra:
Nice module - thanks!

PR #75, opened 2015-07-10
Fix default option in README

stevemao:


contra:
Nice consistency fix - thanks!

PR #77, opened 2015-07-11
Improvements

stevemao:
Make readme more consistent

test on iojs and 0.12	

Expose `options.base` for `dest`
Because the stream returned from `dest` is also readable, it is kinda link a `src` to the next `dest` so the `base` should be configurable just like `src`.


contra:
nice - thanks!

PR #81, opened 2015-07-22
use valid semver range for `engine`

silverwind:
See https://docs.npmjs.com/files/package.json#dependencies and https://github.com/nodejs/io.js/issues/2223.


phated:
For posterity. Isaac said this won't match prerelease versions still...

silverwind:
Yeah, the issue turned out to be `strictEngine` related and the space actually isn't significant. Won't hurt I guess :)

Issue #50, opened 2015-01-06
Write relative symbolic links in `dest`


lexi-lambda:
Currently, symlinks are simply ignored by `dest`—nothing at all is written. This is problematic when trying to write Mac OS X `.framework` bundles, since they heavily depend on proper symlink structure for to manage their resources.

A real example in which this is an issue is in the [`gulp-atom-shell`](https://github.com/joaomoreno/gulp-atom-shell) package. Currently it handles this by writing ZIP files using [`gulp-vinyl-zip`](https://github.com/joaomoreno/gulp-vinyl-zip), which writes the symlinks properly, but this layer of indirection is inconvenient.

It would be nice if relative symlinks could be preserved in the output of `dest`.

contra:
We already added .symlink to vinyl-fs which will be updated in for gulp 4

For a temporary fix you can `npm install vinyl-fs` and use the .symlink function until it is built into the gulp module

lexi-lambda:
@contra Perhaps I'm simply incorrect, but I was under the impression that `symlink` was only for explicitly creating symlinks from any stream. I'm asking for files which are already symlinks in the input to be preserved in the output of the ordinary `dest` function.

limulus:
I’ve run into this issue as well, in my case with trying to add symbolic link support to the [`gulp-untar`](https://www.npmjs.com/package/gulp-untar) package. I’m a bit too new to Gulp to know what the right solution is, but it strikes me that `.dest()` should probably write a symlink if `(file.stat.mode & 0170000) === 0120000)` — the mode for symbolic links [per `stat-mode`](https://github.com/TooTallNate/stat-mode/blob/master/index.js#L18). The destination of the symlink could simply be provided by `file.contents`.

phated:
@contra was this resolved by #85?

contra:
Yep!

PR #85, opened 2015-08-14
Add symlink passthrough with `followSymlinks`

Klowner:
The current behavior is always to resolve symbolic links to actual
files.

This change introduces a `keepsymlinks` option to `src()`. When the
option is set, the destination of each encountered symbolic link is stored
in `file.symlink`. When `dest()` is invoked, the `symlink` value is used
to reconstruct identical symlinks.

It is important to note that the symbolic link path remains completely
unmodified, so care should be taken when sourcing symlinks with absolute
paths.

If found to be sufficient, I'm fairly certain this patch resolves issue #50

### DIFF
--- a/lib/src/resolveSymlinks.js
+++ b/lib/src/resolveSymlinks.js
@@ -4,35 +4,36 @@ var through2 = require('through2');
 var fs = require('graceful-fs');
 var path = require('path');
 
-function resolveSymlinks() {
-  return through2.obj(resolveFile);
-}
+function resolveSymlinks(options) {
 
-// a stat property is exposed on file objects as a (wanted) side effect
-function resolveFile(globFile, enc, cb) {
-  fs.lstat(globFile.path, function (err, stat) {
-    if (err) {
-      return cb(err);
-    }
-
-    globFile.stat = stat;
-
-    if (!stat.isSymbolicLink()) {
-      return cb(null, globFile);
-    }
-
-    fs.realpath(globFile.path, function (err, filePath) {
+  // a stat property is exposed on file objects as a (wanted) side effect
+  function resolveFile(globFile, enc, cb) {
+    fs.lstat(globFile.path, function (err, stat) {
       if (err) {
         return cb(err);
       }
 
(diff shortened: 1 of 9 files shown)


contra:
shouldn't it be the inverse? `followSymlinks: false`

Klowner:
perhaps? The default behavior now is that symlinks are recursively followed into actual paths and therefore they're "lost". So the default behavior would be `followSymlinks: true`, and then using `followSymlinks: false` would activate the code paths in this patch?

contra:
@Klowner yeah, that seems to make the most sense to me

Klowner:
@contra sounds good, I'll change the option name and throw together some unit tests.

contra:
Okay, bump afterwards so I get the notification - gh doesn't notify on new commits inside of PRs

Klowner:
Renamed the option to `followSymlinks` and added unit tests :+1:

contra:
option should be camelcase, then LGTM

lib/src/resolveSymlinks.js
```diff
@@ -4,8 +4,10 @@ var through2 = require('through2');
 var fs = require('graceful-fs');
 var path = require('path');
 
-function resolveSymlinks() {
-  return through2.obj(resolveFile);
+function resolveSymlinks(followSymlinks) {
+  return followSymlinks ?
+    through2.obj(resolveFile) :
```
contra: i think instead of two functions, the resolveFile fn should take in the options object

Klowner:
do you prefer `followSymlinks` or `followSymLinks` ?

I combined the resolveSymlinks functions, although it looks like through2 doesn't permit passing through arguments so `options` is closured in, hopefully that's okay.

contra:
`followSymlinks`

Klowner:
Should be all set then, unless you want it squashed or anything :+1:

contra:
just needs docs

Klowner:
Docs ahoy! :ship:

contra:
Thanks!

Klowner:
Thank you!

PR #99, opened 2015-09-27
add option to strip BOM, default to true - closes #83

### LINKED ISSUE #83: UTF-8 BOM stripping Redux
@sindresorhus @phated @contra @IsaacSanch

https://github.com/wearefractal/vinyl-fs/issues/72 this is awful. I understand your interest in making things generally better, and I like making tooling that fixes legacy problems as much as everyone else. 

But the contents of files are the user's domain. Not the tool's.

We wasted hours trying to find what in our big pipeline was changing our files. _Changing_ content of files is _not_ something that filesystem tools should do. To a filesystem, the contents of a file are _sacred_. Imagine if S3 or Dropbox or HTTP or even TCP subtly "tried to fix files for you". It would be a nightmare. **All cryptographically secure protocols break!**

I urge you to reconsider this. We were using vinyl and vinyl-fs for a bunch of ipfs pipelines, and would like to continue using it. We very much like the viewpoints described here: https://medium.com/@contrahacks/gulp-3828e8126466  But if the vinyl tooling _does not hold it as a **strong guarantee**_ that contents of files can move from A to B without magically changing underneath, we'll have to abandon it for something else. :'(

I think @contra said it best:

> Not adding it just because it was asked for, adding it because it makes sense. We shouldn't be fucking with contents without a way to disable it or we are a really shitty general purpose fs access tool (vinyl-fs)

And I agree. It was disappointing to find this design bug.

phated:
@contra PRed for review

Should be a minor release (2.1.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ fs.src(['*.js', '!b*.js'])
     - `false` will disable writing the file to disk via `.dest()`.
 
   - since - `Date` or `number` if you only want files that have been modified since the time specified.
+
+  - stripBOM - `true` or `false` if you want the BOM to be stripped on UTF-8 encoded files.
+    - Default value is `true`.
+
   - passthrough - `true` or `false` if you want a duplex stream which passes items through and emits globbed files.
     - Default is `false`.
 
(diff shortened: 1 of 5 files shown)


contra:
Awesome, thanks!

PR #100, opened 2015-09-27
allow a function to be used as value for the overwrite option - closes #68

### LINKED ISSUE #68: Overwriting on a per-file basis?
I have been working on a peculiar case: a client has developers working on different editors, and no-one bothered setting their editors / IDEs for tabs and spaces. That client also doesn't really care about style guides and as a result they introduced jsbeautify in their building process: source files are overwritten with their prettified contents. This is of course not editor friendly (causes files to be reloaded, or editors to prompt for reloads) but this is how that client does it.

The build process has been recently converted to Gulp (using Gulp 4), and to try to make the process more editor-friendly, I introduced a piece of pipeline which prettifies file contents, compare the prettified version with its original and only copy if different.

The issue I have is that I bypass gulp.dest() and use fs.writeFile() instead inside that piece of pipeline. I would prefer if I had a piece of pipeline which would prettify and set an `overwrite` flag to true if different, and then pass it to gulp.dest().

[https://github.com/wearefractal/vinyl-fs/blob/master/lib/prepareWrite.js#L15](https://github.com/wearefractal/vinyl-fs/blob/master/lib/prepareWrite.js#L15) could become:

    var overwrite = (file.overwrite !== undefined ? file.overwrite : options.overwrite);
    options.flag = (overwrite ? 'w' : 'wx');

Should I go ahead with a pull request?

phated:
@contra PRed for review

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ _Note:_ UTF-8 BOM will be stripped from all UTF-8 files read with `.src`.
 
   - overwrite - Specify if existing files with the same path should be overwritten or not.
     - Default is `true`, to always overwrite existing files.
+    - Can also be a function that takes in a file and returns `true` or `false`.
 
   - sourcemaps -
     - Default is `null` aka do not write sourcemaps.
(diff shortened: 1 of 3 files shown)


contra:
Nice - I think this should apply to every option in dest as well, maybe for another PR

PR #101, opened 2015-09-28
Issue #1273 (gulp queue): Gulp.src() doesn't follow symlinks

OnkelTem:


Klowner:
does `followSymlinks: false` option not accomplish this?

phated:
As @Klowner stated, I think that options covers this use case

phated:
We have tests for this expected behavior and it is documented.  I don't think we should merge this.

OnkelTem:
As you can see from the code:

```js
function resolveSymlinks(options) {

  // a stat property is exposed on file objects as a (wanted) side effect
  function resolveFile(globFile, enc, cb) {
    fs.lstat(globFile.path, function (err, stat) {
      if (err) {
        return cb(err);
      }

      globFile.stat = stat;

      if (!stat.isSymbolicLink() || !options.followSymlinks) {
        return cb(null, globFile);
      }
      ...
```
if `followSymlinks` is `false` then no symlink resolution is made and then happens what is described here: https://github.com/gulpjs/gulp/issues/1273

OnkelTem:
Also, tests have issues. Give me some time, I'll prepare a PR.

Klowner:
You mention permissions in #1273, is the issue you're having with vinyl keeping symlink permissions or am I misunderstanding?

phated:
I believe this is solved properly in #131 so I'm going to close this.

PR #107, opened 2015-10-01
Send options to through2 via dest(), src(), symlink()

vineethawal:
Ability to send options to through2.
Issue: default value set for highWaterMark in through2 is 16(max buffer length), when buffer length exceeds this default limit all files are not written to destination.


sindresorhus:
:-1: That would only mask your actual problem. The max buffer is 16, true, but it should flush and continue after that, but something is stopping it. You should rather fix the actual problem.

phated:
While I agree with @sindresorhus that the pipeline should be fixed (return the stream, call resume, etc).  I don't think it is wise for us to not allow a user to pass options to through2.  I thought this was already fixed in 2.x though...

phated:
@vineethawal if we are going to support this, it needs to be uniform across the board, in `src`, `dest`, `symlink` (maybe even in glob-stream?).  We aren't __only__ going to support it in `dest`

contra:
@phated Agree, we shouldn't be blocking the users from passing options to the underlying stream. Also, this won't even mask the problem because your plugins will also have a high water mark of 16

vineethawal:
@phated Yes I agree, on it!

phated:
@vineethawal looking good.  Please update the docs to mention that options are passed to through2 and add some tests.  Thanks

vineethawal:
Added Documentation + test cases

vineethawal:
@phated if the PR looks good can you have this merged?

vineethawal:
@phated can we have this merged? waiting to use this feature in my project

phated:
@vineethawal I was reviewing this but we have a bigger problem in issue #127 / #128 that needs to get in ASAP.  I also noticed that only a couple of the changes are being tested.

phated:
@vineethawal looks like I messed up the merge with my docs update, could you rebase and get your docs updates to merge cleanly?  thanks!

vineethawal:
@phated Done!

phated:
Awesome. Thanks! I'm busy during the day today but I will take a look tonight

README.md
```diff
@@ -224,6 +225,10 @@ Type: `Boolean` or `Object`
 
 Default: `undefined` (do not write sourcemaps)
 
+##### other
```
contra: I think we should pull the documentation from the through2 repo instead of linking

phated: I generally agree with surfacing docs but I don't think people should be using these options so I'm good with them not being surfaced.

PR #108, opened 2015-10-10
Fix links, spacing in README

moshen:
Some of the links were broken in the README, I have changed all of them except coveralls.  It doesn't look like there's gulpjs/vinyl-fs coveralls configured.


phated:
Thanks! Been meaning to go through and update links but haven't had time.

Issue #96, opened 2015-09-20
Allow to set file stats on dest()


piranna:
vinyl headers has support for the file stats, but `dest()` is not preserving them. opt field could accept a `stat` field and use it, I'm mostly interested on `mtime` since [download](https://github.com/kevva/download/issues/78) can't be able to preserve it, and source code tarballs fail. This can be done with the [fs.utimes()](https://nodejs.org/api/fs.html#fs_fs_utimes_path_atime_mtime_callback) function.

contra:
Can you clarify a bit more? FYI you can do this in a plugin prior to dest - just modify file.stats

piranna:
Thanks for the advice about the plugin ;-) What I'm asking about is that at https://github.com/contra/vinyl-fs/blob/5b056f66c476ff32687c172d5fe79eaf47b2e80e/lib/dest/writeContents/writeBuffer.js#L6-L11 and https://github.com/contra/vinyl-fs/blob/5b056f66c476ff32687c172d5fe79eaf47b2e80e/lib/dest/writeContents/writeStream.js#L7-L12 file is created with current `mtime` instead of the one defined on `file.stats`. `fs.writeFile()` and `fs.createWriteStream()` don't allow to set the stats or the time on their arguments, but this can be set with the `fs.utimes()` function, just call it after the file is created.

contra:
@piranna Interesting... could you explain your use case a bit more? You might want to try using symlinks instead of having gulp copy the files over - use gulp.symlink instead of gulp.dest

phated:
@contra the referenced `download` issue has more reasoning why they need the `mtime` from stats.  It seems that some files can't be unzipped if their created time is different than the original tar

piranna:
yes, @phated is (almost) true, some times source code tarballs can't be compiled if their time is different of the original ones since `make` think they have been changed, that's why it's needed to preserve the original time. vynil headers use already the current time by default on the `stat` entry, so it's ok to set them always after the files are created.

piranna:
I've created a [pull-request](https://github.com/gulpjs/vinyl-fs/pull/110) to fix this.

phated:
Published this change as 2.2.0

piranna:
Thank you :-)

PR #110, opened 2015-10-14
Set utimes of files if they are provided

piranna:
This fixes issue #96

### DIFF
--- a/lib/dest/writeContents/utils.js
+++ b/lib/dest/writeContents/utils.js
@@ -1,0 +1,15 @@
+'use strict';
+
+var fs = require('graceful-fs');
+
+function utimes(writePath, stat, cb) {
+  if (stat.mtime) {
+    stat.atime = stat.atime || new Date();
+
+    return fs.utimes(writePath, stat.atime, stat.mtime, cb);
+  }
+
+  cb();
+}
+
+exports.utimes = utimes;
--- a/lib/dest/writeContents/writeBuffer.js
+++ b/lib/dest/writeContents/writeBuffer.js
@@ -2,7 +2,7 @@
 
 var fs = require('graceful-fs');
 
-var utimes = require('./utils').utimes;
+var utimes = require('../../utimes');
 
 function writeBuffer(writePath, file, cb) {
   var stat = file.stat;
@@ -12,8 +12,7 @@ function writeBuffer(writePath, file, cb) {
     flag: file.flag
   };
 
-  fs.writeFile(writePath, file.contents, opt, function(error)
-  {
+  fs.writeFile(writePath, file.contents, opt, function(error) {
     if (error) {
       return cb(error);
     }
--- a/lib/dest/writeContents/writeBuffer.js
+++ b/lib/dest/writeContents/writeBuffer.js
@@ -3,12 +3,27 @@
 var fs = require('graceful-fs');
 
 function writeBuffer(writePath, file, cb) {
+  var stat = file.stat;
+
   var opt = {
-    mode: file.stat.mode,
+    mode: stat.mode,
     flag: file.flag
   };
 
-  fs.writeFile(writePath, file.contents, opt, cb);
+  fs.writeFile(writePath, file.contents, opt, function(error)
+  {
+    if(error)
+    {
+      return cb(error);
+    }
+
+    if(stat.atime && stat.mtime)
+    {
+      return fs.utimes(writePath, stat.atime, stat.mtime, cb);
+    }
+
+    cb();
+  });
 }
 
 module.exports = writeBuffer;
--- a/lib/dest/writeContents/writeBuffer.js
+++ b/lib/dest/writeContents/writeBuffer.js
@@ -2,6 +2,8 @@
 
 var fs = require('graceful-fs');
 
+var utimes = require('./utils').utimes;
+
 function writeBuffer(writePath, file, cb) {
   var stat = file.stat;
 
@@ -12,19 +14,11 @@ function writeBuffer(writePath, file, cb) {
 
   fs.writeFile(writePath, file.contents, opt, function(error)
   {
-    if(error)
-    {
+    if (error) {
       return cb(error);
     }
 
-    if(stat.mtime)
-    {
-      stat.atime = stat.atime || new Date();
-
-      return fs.utimes(writePath, stat.atime, stat.mtime, cb);
-    }
-
-    cb();
+    utimes(writePath, stat, cb);
   });
 }
 
--- a/lib/dest/writeContents/writeBuffer.js
+++ b/lib/dest/writeContents/writeBuffer.js
@@ -17,8 +17,10 @@ function writeBuffer(writePath, file, cb) {
       return cb(error);
     }
 
-    if(stat.atime && stat.mtime)
+    if(stat.mtime)
     {
+      stat.atime = stat.atime || new Date();
+
       return fs.utimes(writePath, stat.atime, stat.mtime, cb);
     }
 
--- a/lib/dest/writeContents/writeStream.js
+++ b/lib/dest/writeContents/writeStream.js
@@ -4,8 +4,10 @@ var streamFile = require('../../src/getContents/streamFile');
 var fs = require('graceful-fs');
 
 function writeStream(writePath, file, cb) {
+  var stat = file.stat;
+
   var opt = {
-    mode: file.stat.mode,
+    mode: stat.mode,
     flag: file.flag
   };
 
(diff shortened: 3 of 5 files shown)


phated:
I'm a :+1: on this.  Needs to be updated to match our code style though (braces on same line, etc) and needs tests.  Maybe the implementation should be a utility because it is exactly the same in both places.  I'd hate for those to get out of sync.

piranna:
I've updated the style and made an utility function. I've also done a little test but it fails, the written mtime has lost the second decimals:

```
  1) dest stream should write file mtime using the vinyl mtime:

      Uncaught AssertionError: expected 2015-10-14 17:32:17.000 +0200 to be 2015-10-14 17:32:17.300 +0200
      + expected - actual

      -[Date: 2015-10-14T15:32:17.000Z]
      +[Date: 2015-10-14T15:32:17.300Z]
```

I don't know up to what point this should be expected... :-/ What do you think?

piranna:
Ok, [somewhat expected](http://linux.die.net/man/2/utimes):

> The utime() system call allows specification of timestamps with a resolution of 1 second.
>
> The utimes() system call is similar, but the times argument refers to an array rather than a structure. The elements of this array are timeval structures, which allow a precision of 1 microsecond for specifying timestamps.

So since Node.js [use utime()](https://github.com/nodejs/node/blob/6fff47ffacfe663efeb0d31ebd700a65bf5521ba/deps/uv/src/unix/fs.c#L613) and only have 1 second resolution, I'll change the test to consider this.

piranna:
Fixed, now adjusting the test to 1 second resolution it works :-)

lib/dest/writeContents/writeBuffer.js
```diff
@@ -2,13 +2,24 @@
...
+  var stat = file.stat;
+
   var opt = {
-    mode: file.stat.mode,
+    mode: stat.mode,
     flag: file.flag
   };
 
-  fs.writeFile(writePath, file.contents, opt, cb);
+  fs.writeFile(writePath, file.contents, opt, function(error)
```
contra: you should use the async module here instead of nesting

lib/dest/writeContents/writeStream.js
```diff
@@ -18,7 +22,14 @@ function writeStream(writePath, file, cb) {
   file.contents.pipe(outStream);
 
   function success() {
-    streamFile(file, {}, complete);
+    streamFile(file, {}, function(error)
```
contra: use async

lib/dest/writeContents/writeBuffer.js
phated: we don't have a dep on async and i don't think we should add it

contra:
Nice! I think we should use async for the control flow otherwise looks good - thanks!

lib/dest/writeContents/writeStream.js
phated: we don't have a dep on async and i don't think we should add it

piranna:
So is needed async or not? If so, it should be used everywhere, so I think is not only a thing of this pull-request...

contra:
@piranna I'd like to use it everywhere, I'm assuming that phated is going to call it "bloated" - we can leave that our of this PR but I think it would be useful to clean up the control flow eventually, maybe with some smaller more focused modules.

piranna:
Ah, ok then :-)

I don't think async is bloated, at least it don't use compiled modules and it's very stable and used by everybody so it can be shared by several modules inside a project... But I must to admit that's big and difficult to debug and sometimes you only need one or two functions... :-/

lib/dest/writeContents/utils.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+var fs = require('graceful-fs');
+
+function utimes(writePath, stat, cb) {
+  if (stat.mtime) {
+    stat.atime = stat.atime || new Date();
```
phated: this is mutating the `atime` on the file object.  I don't think we want that, correct?

lib/dest/writeContents/utils.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
phated: this file can be `./lib/utimes.js` similar to prepareWrite.js

lib/dest/writeContents/utils.js
```diff
@@ -0,0 +1,15 @@
...
+  if (stat.mtime) {
+    stat.atime = stat.atime || new Date();
+
+    return fs.utimes(writePath, stat.atime, stat.mtime, cb);
+  }
+
+  cb();
+}
+
+exports.utimes = utimes;
```
phated: just module.exports the function instead of assigning it to a property

lib/dest/writeContents/writeBuffer.js
```diff
@@ -2,13 +2,24 @@
...
+  {
```
phated: move this brace to the same line as fs.writeFile

lib/dest/writeContents/writeStream.js
```diff
@@ -18,7 +22,14 @@ function writeStream(writePath, file, cb) {
...
+    {
```
phated: move this brace to streamFile

test/dest.js
```diff
@@ -574,6 +574,46 @@ describe('dest stream', function() {
     stream.end();
   });
 
+  it('should write file mtime using the vinyl mtime', function(done) {
```
phated: can you add a test for atime?  both setting it and getting a defaulted one

phated: can you add a test for not mutating the state property of a vinyl object?

lib/dest/writeContents/utils.js
piranna: This add a new `atime` equals to current time if no one is provided. This is done this way because `fs.utimes()` need both `atime` and `mtime`.

lib/dest/writeContents/utils.js
piranna: I did it this way in case more util functions could be added later... :-/

lib/dest/writeContents/utils.js
piranna: Done.

lib/dest/writeContents/writeBuffer.js
piranna: Done.

lib/dest/writeContents/writeStream.js
piranna: Done.

test/dest.js
piranna: All changes and tests are done. Could you be able to review it?

lib/dest/writeContents/utils.js
phated: yeah, but we don't want to mutate the underlying vinyl object, which is what this is doing.  If you instead do `var atime = stat.atime || new Date()`, you won't be mutating the underlying object.  This needs a test too

lib/utimes.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+var fs = require('graceful-fs');
+
+function utimes(writePath, stat, cb) {
+  if (stat.mtime) {
+    stat.atime = stat.atime || new Date();
```
phated: this is still mutating the underlying vinyl object.  If you instead do `var atime = stat.atime || new Date()`, you won't be mutating the underlying object.  This needs a test too.

piranna: Ok! I didn't get it before, good point. Fixed and added tests, although they are failing because someone is adding an undefined `mode`... :-/ I've done it on a different commit, so it could be ignored on this pull-request if needed and cherry-picked on another branch to search who is adding the `mode`.

phated:
@piranna do we need to make changes on vinyl to guarantee that mode is an object?

phated:
nevermind the previous comment.  It seems that vinyl mutates the file object at https://github.com/gulpjs/vinyl-fs/blob/master/lib/prepareWrite.js#L50

phated:
@contra what do you think about changing https://github.com/gulpjs/vinyl-fs/blob/master/lib/prepareWrite.js#L50 to only set the mode if we actually have something for mode?  That way we won't be adding a property with value of undefined

piranna:
> @contra what do you think about changing https://github.com/gulpjs/vinyl-fs/blob/master/lib/prepareWrite.js#L50 to only set the mode if we actually have something for mode? That way we won't be adding a property with value of undefined

:+1: Totally makes sense.

contra:
@phated yep makes sense

phated:
Interesting change was made to the fs.Stat object in node 0.12;  They now have every property you could imagine on the instance.  I think we actually need to normalize this for older node.

phated:
Thanks for this.  In light of my findings, I just tweaked the tests in https://github.com/gulpjs/vinyl-fs/commit/039d68920cc5bde6f0f12375daee2713260a1ecf

phated:
Published as 2.2.0

piranna:
Cool, thanks for updating my commit. Just a question: is not redundant check first for the property and later for it's value? If the value is not there, it will be `undefined` anyway...

phated:
@piranna `should` explodes if the property doesn't exist (since it modifies each property) so I couldn't just check for the values to be the same.

piranna:
> @piranna should explodes if the property doesn't exist (since it modifies each property) so I couldn't just check for the values to be the same.

Ah, ok, that makes sense. Thanks :-)

Issue #113, opened 2015-10-16
2.2.0 writes negative timestamps


grassick:
2.2.0 writes negative timestamps under some circumstances:

```
-rw-r--r--  1 clayton clayton 1238 Oct 15 11:49 index_map.html
-rw-r--r--  1 clayton clayton 1192 -9223372036854775808 manifest.appcache
-rw-r--r--  1 clayton clayton  196 Feb 20  2015 Readme.md
-rw-r--r--  1 clayton clayton   25 Feb 20  2015 robots.txt
```

2.1.1 did not. I see recent commits with mtime changes. This is in node 0.12.7.

grassick:
2.2.0 also breaks watchify by causing `end` events to not fire anymore in gulp. Still investigating that problem.

phated:
Can you give me a reproduction case? We added a new feature in 2.2.0 to set mtime and atime if set on a vinyl object.

phated:
cc'ing @piranna to look into this also

piranna:
Maybe the original file has badly set the mtime of atime fields? Since now we are using them, maybe it have shown a hidden bug elsewhere...

phated:
@piranna I wonder if this could come up because node 0.12 added properties to the fs.Stats object and made the atime, mtime, etc Date objects and passing undefined to `new Date()` causes an invalid date.  Thoughts?

phated:
@piranna maybe we need a check like http://stackoverflow.com/a/10589791 inside the utimes helper?

piranna:
That time converted to hexadecimal is `FFFFFFFFFA80A008`... A little bit big, isn't it? I'm not sure if it's a special value or something...


> @piranna maybe we need a check like http://stackoverflow.com/a/10589791 inside the utimes helper?

I would not hurt in any case... What should do in case of a bad value, ignore it or raise an error?

phated:
@piranna due to node 0.12 having NaN dates in my quick tests (run `new fs.Stats()` in the repl), I think we have to just ignore them.

piranna:
I've just done a [pull-request](https://github.com/gulpjs/vinyl-fs/issues/114) to fix this. It ignore the atime and mtime fields if any of both is invalid.

phated:
Sorry for the delay.  I wanted to make sure our tests had a reproduction case and I could verify we fixed this before was published.  I've just published this as 2.2.1 - please let me know how this works out for you.

PR #114, opened 2015-10-16
[Fix] Ignore atime and mtime if they are invalid

piranna:
Fixes issue #113

### DIFF
--- a/lib/utimes.js
+++ b/lib/utimes.js
@@ -2,11 +2,20 @@
 
 var fs = require('graceful-fs');
 
+// http://stackoverflow.com/a/10589791/586382
+function validDate(date) {
+  return date instanceof Date && !isNaN(date.valueOf());
+}
+
 function utimes(writePath, stat, cb) {
-  if (stat.mtime) {
+  var mtime = stat.mtime;
+
+  if (mtime) {
     var atime = stat.atime || new Date();
 
-    return fs.utimes(writePath, atime, stat.mtime, cb);
+    if(validDate(atime) && validDate(mtime)) {
+      return fs.utimes(writePath, atime, mtime, cb);
+    }
   }
 
   cb();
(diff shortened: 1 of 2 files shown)


phated:
@piranna the callback isn't called if we have mtime but it is invalid.  also, can you add a test or two for this?  thanks!

piranna:
I've just added the tests.

phated:
@piranna thanks. is there a way you can think of to test the file didn't receive an invalid timestamp? Maybe check it is within some range of Date.now()?  I'm not too sure here.  It just seems like testing we didn't change the vinyl object is the wrong thing to be testing.

lib/utimes.js
```diff
@@ -2,11 +2,20 @@
...
+// http://stackoverflow.com/a/10589791/586382
+function validDate(date) {
+  return date instanceof Date && !isNaN(date.valueOf());
+}
+
 function utimes(writePath, stat, cb) {
-  if (stat.mtime) {
+  var mtime = stat.mtime;
+
+  if (validDate(mtime)) {
```
contra: `if (validDate(mtime) && validDate(atime)) {

contra:
We should land this ASAP - the latest version on npm (which is being used by 4.0) is outputting broken files and people are pretty upset about that.

phated:
@contra we need a better way to test this.  feedback welcome

phated:
@piranna can you please fix the tests ASAP so we can get this published?  Thanks

phated:
I believe I fixed the tests to verify that they aren't setting invalid values over on the [invalid-utimes branch](https://github.com/gulpjs/vinyl-fs/tree/invalid-utimes)

phated:
Deleted branch and merged to master.  Waiting for travis to run and bumping a patch release.

phated:
Published as 2.2.1

piranna:
Sorry for the delay, I've been out due to holydays, that's why I tried to had it ready the last friday. I'm happy this could be fixed.

PR #115, opened 2015-10-22
Add ESLint and JSCS to repo

pdehaan:
```sh
$ npm run lint

> vinyl-fs@2.2.1 lint /Users/pdehaan/dev/github/vinyl-fs
> jshint *.js lib/ && jscs *.js lib/

$ echo $?
0
```

Bonus points if we want to add linting to the tests/ directory as well.


phated:
@pdehaan could you update this to merge cleanly?

pdehaan:
Rebased. Works locally, apart from ~50 ESLint warnings (!) and one odd `npm test` failure:
```
  74 passing (7s)
  1 failing

  1) dest stream should write file mtime using the vinyl mtime:
     ReferenceError: buffered is not defined
      at Context.<anonymous> (test/dest.js:690:45)
```

pdehaan:
Looks like all the ESLint warnings are one of these:
```
   422:59   warning  This function has too many statements (16). Maximum allowed is 15  max-statements
   427:9    warning  "someVariable" is defined but never used                            no-unused-vars
```

Assuming we don't want to refactor ~20 methods, we can probably disable that ESLint rule for this repo and live with the shame. We can probably evaluate the other 30 `no-unused-vars` violations in a future PR or somethin'.

contra:
@pdehaan If you have time to fix the warnings that would be super helpful

phated:
Awesome. Thanks! I'm busy during the day today but will take a look tonight

pdehaan:
ESLint warnings suppressed. :christmas_tree:

.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "gulp",
+
+  "rules": {
+    "max-statements": 0
```
pdehaan: SHAME!

index.js
```diff
@@ -3,5 +3,5 @@
 module.exports = {
   src: require('./lib/src'),
   dest: require('./lib/dest'),
-  symlink: require('./lib/symlink')
+  symlink: require('./lib/symlink'),
```
contra: trailing comma? when did that become something people did

lib/dest/writeContents/index.js
```diff
@@ -60,17 +60,16 @@ function writeContents(writePath, file, cb) {
   }
 
   function isErrorFatal(err) {
+    var isFatal = true;
```
contra: which eslint rule is this?

index.js
pdehaan: https://github.com/gulpjs/gulp/blob/master/bin/gulp.js#L25

I think it's silly, but your other repo's use it. Can't remember if it was via ESLint or jacks though.

lib/dest/writeContents/index.js
pdehaan: This was some weird rule about an `else` after a return statement. I don't remember seeing this warning before though in the other repo's. Seems like a funny rule since your code was legit, but I can see how it could get confusing in other complex cases.

index.js
phated: pretty sure it is from node-style-guide and wasn't worth an override.

.eslintrc
phated: I'm going to remove this to keep getting the warnings so we can fix it in the future

lib/dest/writeContents/index.js
phated: yeah, the rule is "no else after return" i think - might be a different way to do this

index.js
pdehaan: Ah yes! [**jscs-dev/node-jscs** /presets/node-style-guide.json:41-43](https://github.com/jscs-dev/node-jscs/blob/1afee411b8210ef4a7a094e4997d01003ebbd43f/presets/node-style-guide.json#L41-L43):

```js
    "requireTrailingComma": {
        "ignoreSingleLine": true
    },
```

lib/dest/writeContents/index.js
pdehaan: I think it _may_ be this: http://jscs.info/rule/requireEarlyReturn

phated: I changed it at https://github.com/gulpjs/vinyl-fs/commit/00ece7ef6b794fd1999695fde2d88bba8d69f5d0

Issue #120, opened 2015-10-30
vfs-dest doesnt end?


davidchase:
Maybe i missing something so i apologize in advanced... i know `vfs.dest` returns a readable/writable stream... so is up to the user to end that stream?

for example if i simply do `vfs.src.pipe(vsf.dest)` and the source contains 17+ files the destination folder will only have 16 files since thats the limit of the highWaterMark, 

however if i pipe from `dest` like so `vfs.src.pipe(vsf.dest).pipe(...)` or attach a `data` listener then i will get all of the files from the source to destination since i drain and keep the flow going...

so that additional pipe or adding a `data` listener to simply copy files in my basic example feels a bit odd but i can understand if thats the intent to have the user of `vinyl-fs` to choose what happens with the stream.

thanks
-david

phated:
dest should be pulling files from the stream. It always has, the only time you should have to drain is if you don't have a dest or some other stream pulling on the end of the pipeline

davidchase:
@phated thanks for the quick reply but i guess im not exactly following..

lets say i have 17 files inside `client/src/js` and i want to move them to `client/dest/js` for sake of simplicity i have no transforms its essentially a "copy/paste":

if i do `vfs.src('./client/src/*.js').pipe(vfs.dest('client/dest/js'))` only 16 of those 17 files would move from the source directory to the destination directory.

unless i add some `pipe` after the `dest`, is that a correct intuition ?

phated:
No, the dest method was written in a way that empties the stream so you won't pause at 16 files

davidchase:
but only 16 out of 17 files transfer to the `dest` folder something is not working...

with the gulp and older `vinyl-fs` which is `0.3.0` it transfers all files with gulp and a newer `vinyl-fs` at `2.2.0` it does **not** transfer all of the files

according to this commit https://github.com/gulpjs/vinyl-fs/commit/99db7de8f0485c5ca158c79126b53b8fdab9d821#diff-ebdf4495a061217916345559e7783922L19 

it used to do `stream.resume()`

phated:
@davidchase thanks for tracking that down.  That's what I was trying to understand.  Sorry for being confusing, every response was from my phone.  This is something we need to fix.

davidchase:
[...]

PR #123, opened 2015-11-02
Re-add stream resume in dest

Klowner:
It appears as though stream.resume() was removed in commit 99db7de8f0,
this causes streams to prematurely end unless dest() is followed by
something providing a stream drain.

This change re-adds stream.resume(), causing dest() to consume the
entire incoming stream regardless.

I believe this fixes issue #120


Klowner:
This may be a crap solution.

Klowner:
This is definitely a crap solution, I'm gonna close this

:dancers: :dancers: :dancers:

PR #126, opened 2015-11-05
read from the stream once we finish processing a file, this avoids highWaterMark being hit - add test

### LINKED ISSUE #120: vfs-dest doesnt end?

phated:
@contra  I think I solved #120 with this.  Can you take a look? I made the test and it was failing and then I made it pass.  @davidchase could you also take a look at this?

Closes #120

### DIFF
--- a/lib/dest/index.js
+++ b/lib/dest/index.js
@@ -3,6 +3,7 @@
 var through2 = require('through2');
 var sourcemaps = require('gulp-sourcemaps');
 var duplexify = require('duplexify');
+var sink = require('../sink');
 var prepareWrite = require('../prepareWrite');
 var writeContents = require('./writeContents');
 
@@ -21,7 +22,12 @@ function dest(outFolder, opt) {
   }
 
   var saveStream = through2.obj(saveFile);
+
   if (!opt.sourcemaps) {
+    // Sink the save stream to start flowing
+    // Do this on nextTick, it will flow at slowest speed of piped streams
+    process.nextTick(sink(saveStream));
+
     return saveStream;
   }
 
@@ -33,6 +39,10 @@ function dest(outFolder, opt) {
   var mapStream = sourcemaps.write(sourcemapOpt.path, sourcemapOpt);
   var outputStream = duplexify.obj(mapStream, saveStream);
   mapStream.pipe(saveStream);
+
+  // Sink the output stream to start flowing
+  // Do this on nextTick, it will flow at slowest speed of piped streams
+  process.nextTick(sink(outputStream));
 
   return outputStream;
 }
--- a/lib/sink.js
+++ b/lib/sink.js
@@ -1,0 +1,16 @@
+'use strict';
+
+var Writable = require('readable-stream/writable');
+
+function sink(stream) {
+  var sinkStream = new Writable({
+    objectMode: true,
+    write: function(file, enc, cb) { cb(); }
+  });
+
+  return function() {
+    stream.pipe(sinkStream);
+  };
+}
+
+module.exports = sink;
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "merge-stream": "^1.0.0",
     "mkdirp": "^0.5.0",
     "object-assign": "^4.0.0",
+    "readable-stream": "^2.0.4",
     "strip-bom": "^2.0.0",
     "strip-bom-stream": "^1.0.0",
     "through2": "^2.0.0",
@@ -27,6 +28,7 @@
   },
   "devDependencies": {
     "buffer-equal": "^0.0.1",
+    "del": "^2.2.0",
     "istanbul": "^0.3.0",
     "istanbul-coveralls": "^1.0.1",
     "jshint": "^2.4.1",
--- a/test/dest.js
+++ b/test/dest.js
@@ -8,7 +8,8 @@ var vfs = require('../');
 
 var path = require('path');
 var fs = require('graceful-fs');
-var rimraf = require('rimraf');
+var del = require('del');
+var Writeable = require('readable-stream/writable');
 
 var bufEqual = require('buffer-equal');
 var through = require('through2');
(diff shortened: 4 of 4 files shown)


lib/dest/index.js
```diff
@@ -12,11 +12,21 @@ function dest(outFolder, opt) {
   }
 
   function saveFile(file, enc, cb) {
+    /* jshint validthis: true */
+    var stream = this;
+
+    function complete(err, file) {
+      cb(err, file);
+      if (!err) {
```
davidchase: seems very similar to my proposition in #120 but can you enlighten me how the error propagates or not in this situation?

is dealing with the stream being `done` or rather flushed as I proposed not ideal ?

phated: the return value from push isn't a done flag (or that's what I think I found in all my digging).  I tried implementing your suggestion in #120 but it didn't seem to work.  The `cb` is being called on the line above with the `err` and `file` objects.  If an error occurs and is passed to that, the `error` event should be emitted on the stream and the pipeline torn down.  I check to see if there wasn't an error before reading because I don't know if reading from a torn down stream blows up or not.

davidchase: Hmm, so http://bit.ly/1kyt00f says that readable will return a boolean which will say if more data should be pushed into the buffer which is controlled by the highWaterMark http://bit.ly/1kytDGX then as it hits false in the case of theres no piping happening we can then `stream.read` to pull chunks again from the buffer http://bit.ly/1kytTpp. When `piping` from a destination `stream.read` never gets called because the `.pipe` handles the draining of the readable part http://bit.ly/1kyvnjr

I've tried my way with a pipe from the `dest` and not piping essentially `finish`ing the write part and it works..

phated: @davidchase you are correct that the flag means more data can be pushed; however, that has no bearing on if we can read more.  As soon as we are process the file object and pass it on (calling the callback with err & file), we are able to read another and process that.  As far as I am aware, backpressure is handled by the reading stream and because we weren't reading or resuming, the beginning of our stream was clogging the pipeline.  As you stated, the piping of dest to somewhere else will drain it, so the read will essentially nothing in that case (I think).

davidchase: > ; however, that has no bearing on if we can read more

Ahh i guess I misunderstood
>  _read should start pushing that data into the read queue by calling this.push(dataChunk). _read should continue reading from the resource and pushing data until push returns false, at which point it should stop reading from the resource...

https://nodejs.org/api/stream.html#stream_readable_read_size_1


> so the read will essentially nothing in that case (I think).

yeah looks like it returns null when `pipe` from `dest`

contra:
@phated ping me when you need a look

phated:
@contra can you review the comments between @davidchase and I and help us figure out the correct solution here?  My stream-fu seems to be weak.

lib/dest/index.js
contra: @davidchase After consulting the stream docs that seems like it should work. We should make sure our stream implementation is in line with the latest docs

contra:
Just commented. Also: can we avoid the cruft of having a ton of fixture files and make those dummy files in code? This will also make it easier to test different #s later on down the road

davidchase:
any luck with this one?

phated:
I reached out to @chrisdickinson and @nrn on IRC but I don't think either was available to give any guidance.

nrn:
Sorry, checked it out but I haven't dealt with the subtleties around this enough to have any insight. Good luck!

phated:
@davidchase it seems your suggestion passes the test I added so I switched to that.  Also updated as per @contra's suggestions and rebased.

Let me know what you think

phated:
@davidchase can you think of a test for when `stream.push()` will return false?  I'd like to add that to the suite if we can come up with one.

davidchase:
@phated i think the only time `stream.push` would return false is in the case when we would be using the  internal buffer which maxes out at the highWaterMark (16 files/objects) then it cant push anymore data and it returns false.

But if we are using `stream.read()` we are effectively bypassing the internal buffer altogether which makes me wonder if that is ideal???

phated:
@davidchase I thought `this.read()` would pull from the previous streams' buffer or queue up a read if there is nothing in it, not empty our buffer.  And I thought `this.push()` pushed things onto our buffer (that the next stream pulls from) and returned false if we were at the max (however, I couldn't figure out how to write a test that showed that part).

phated:
k, went through this with chris and he explained what is happening in the current implementation and the best way to fix this.  I will update this PR and try to add another test soon.

davidchase:
Sounds good :+1:

phated:
@contra @davidchase I've updated with @chrisdickinson's suggestion and added another test.  Can you take a look again.

davidchase:
@phated interesting can you explain a little bit of whats happening... are you taking the through stream and just piping it a sink that just calls the `cb` ? 

this seems very similar to the [stream-exhaust](https://github.com/chrisdickinson/stream-exhaust) approach, is that a correct assumption ?

contra:
Review, LGTM - thanks @phated 🙏

phated:
@davidchase I believe a pipeline must end in a writeable stream to function correctly, so this was the suggestion from @chrisdickinson.  I asked how it was different than just calling `.resume()` to which he replied

> it should be a bit more durable than `.resume` in that it will count as one of the downstream pipes, so even if another dest stream is unpiped the saveStream will continue flowing

and then further on in the discussion, he said the following, which I believe sums up the change nicely:

> to be on the safe side I'd probably just pipe `saveStream` to an infinitely fast writable sink — that way it will only go as fast as the slowest stream it is piped to, but if it is unpiped it will continue to function

You are correct that it is similar to the stream-exhaust approach (I even asked chris about that) and I considered using it as a dependency, but since we are in control of the streams, we shouldn't need the fallback to the `.resume` call.  The stream-exhaust module is still used deep inside undertaker incase people return a stream that ends with a transform stream that isn't gulp.dest but this change will make vinyl-fs stream correctly when not inside a gulp task.

phated:
:shipit:

asgoth:
Any idea when this fix will be in a new version of vinyl-fs?

phated:
Soon. I'm dealing with some personal stuff right now but hoping to dive back in soon.

Issue #127, opened 2015-11-06
utime/futimes can only be used on files that you own


phated:
This was pointed out in https://github.com/gulpjs/gulp/issues/1376 and it is documented in the manpage of utime (http://man7.org/linux/man-pages/man2/utimes.2.html).  We should check the file owner before trying to call futimes on it.

@piranna @erikkemperman

piranna:
sh*t... I know it's not my fault, but I must to admit I felt guilty... :-P Yeah, I'll check the owner before trying to change the utimes, but as you have said on the other issue, this should *not* be a problem since the files *should* be created with the correct owner... :-/

ounos:
Thank you. :+1:

phated:
@piranna it seems that we inherit the owner and stuff from the `src` files (or so it is documented).

piranna:
> @piranna it seems that we inherit the owner and stuff from the src files (or so it is documented).

Via the `uid` and `gid` fields, isn't it? Makes sense... Anyway this check should be done to prevent failures like the one on the gulp issue, isn't it?

phated:
@piranna yeah, we need to check the owner to avoid that issue.

piranna:
By the way, why the change from `utime` to `futimes`? Have they fixed the sub-second precision only on the `fd*` functions?

phated:
@piranna yep, because it uses futimes underneath which has sub-second precision on filesystems that support it.

erikkemperman:
@piranna There was talk [here](https://github.com/joyent/libuv/issues/1371) about making `fs.utimes` use the `utimes` system call in stead of `utime`. As the manpage linked above show, one supports sub-second precision and the other doesn't. So yeah, that's why I proposed that change.

I think this ownership issue is orthogonal to `utimes` vs `utime` though.

piranna:
> I think this ownership issue is orthogonal to utimes vs utime though.

Yeah, definitely, I was just curious about that thing. Absolutely +1 to use `utimes` :-) Only I was a bit surprised about the usage of file description instead of path functions, no more.

erikkemperman:
It would be good if node would use the `utimes` syscall consistently. But that issue is already a year old and has no follow-up as far as I can tell. And even it were fixed today, it would take some time before a lot of node users actually benefit from it. So for the moment we're going to have to deal with filedescriptors where it would otherwise not be necessary.

piranna:
I've found the change was easy:

```Javascript
[...]

PR #128, opened 2015-11-06
[fix] `futimes` only works for owned files with write access

piranna:
This fixes issue #127


phated:
@piranna I think we need to `lstat` the output file before opening and skip over that whole part if we aren't the owner.  If we don't do the `lstat`, we modify the atime on the file (due to it being accessed with `open`).  If we do that first, we can avoid swallowing the error upon EPERM.

piranna:
Are you talking about doing the ownership check first instead of capturing
the exception?

contra:
Just a note: keep in mind that more fs calls = way slower. dest was only 1 fs call - now it seems to be ballooning to add new features. Performant and minimal should be the default, features should be enabled as needed.

phated:
src has a lot of calls to fs methods, so I don't think dest will be our bottleneck.  However, maybe we need to consider making things like these extra methods on the vinyl-fs adapter.  e.g. - `vfs.utimes()` that generates a stream of utime updates on the vinyl files passed through.  /cc @contra

piranna:
Long-term idea: what about make dest to know when we are trying to create a
new file and when to update it? Not also would be useful to diferenciate it
on this situations, but also would make it more secure by using 'wx' and
'r+' open modes by default instead of just 'w'...

contra:
@phated yeah, i'm more in favor of alternate functions (vs. cramming everything into dest)

piranna:
I've updated the pull-request to check if we are the owners of the file instead of capturing the `EPERM` exception. The code maybe it's cleaner and could be used as basis to solve the problems with `chown` or `chmod`, but it's obvious it less performant...

erikkemperman:
@phated What would be the advantage of doing up to two filesystem calls (`fs.lstat` and `fs.futimes`) instead of just one with a catch?

@contra I understand the argument to keep things like `futimes` (and possibly `chown` and `chmod`) out of `dest` to keep the basic use case quick... But couldn't an argument also be made to keep the file on disk and the vinyl object as similar as possible, including metadata like times, mode and owner? I mean by default, not requiring extra vinyl-fs methods? I tried to make this case [here](https://github.com/gulpjs/vinyl-fs/issues/127#issuecomment-155196878).

piranna:
Maybe it could be used a flag to set if instead throw an error as now happens, swallow it and update the vinyl object with the content from the filesystem, this way we don't break current functionality but also allow to give an oportunity to that methods that require higher priviledges.

phated:
@piranna I think this feature needs to be behind an option flag or in a separate method.

piranna:
> @piranna I think this feature needs to be behind an option flag or in a separate method.

Are you talking about the swallow of errors of priviledged methods like chmod or futimes and update of the vinyl object, isn't it? I agree on the swallow of the errors as a flag, but maybe the update of the vinyl object should have its own flag, isn't it? Do you want I implement it?

phated:
@piranna We shouldn't swallow any errors but as @contra said, we shouldn't be making these `fs` calls for every file.  Either we make a method (`vfs.touch()` maybe?) or hide the fs.futimes calls behind an option.

piranna:
I suppose that `vfs.touch()``would apply all this priviledged functionality at once (uid, gid, utimes, mode...), isn't it?

phated:
@piranna no, touch would be equivalent to https://en.wikipedia.org/wiki/Touch_(Unix) (doesn't modify uid, guid, modes)

piranna:
Do you propose then to have one method for each of the entries? I would prefer a `vinylfs.stat()` method that apply all the stat values at once (uid, guid, mode, futimes, whatever) from the vinyl object to the dest file... If you are interested in one of them, probably you are in fact interested in all of them.

phated:
I think that could be beneficial but I don't like the idea of calling it `vfs.stat()` because the node API has a stat method that just reads them (no modification).

Also, having a singular method for those would make the implementation very complex.

I've always viewed the vinyl adapter spec as having 2 methods (src & dest), especially now that we removed watch.  However, we also have added symlink to vinyl-fs but that method doesn't make sense for other transports (such as vinyl-http).  Maybe symlink, touch, chmod, etc are outgrowing the base spec and could become an addon library.  Thoughts?  /cc @contra

phated:
^ the addon library could still be included with gulp 4

piranna:
> I don't like the idea of calling it vfs.stat()

I though first about `applyStat()`, but wanted to make it simpler, like `dest()`- The name is not (so) important, just what it does.


> Also, having a singular method for those would make the implementation very complex.

Not really, they could be done easily and in a generic way with try-catch, for example.

erikkemperman:
@phated @piranna 
Well, if we want to stick with Node `fs` nomenclature then we'd wind up with `vfs.utimes`, `vfs.chmod` and `vfs.chown`. Or maybe better, stick with what people know from the commandline and make the first one `vfs.touch`.

But because of the sub-second quirk we ran into earlier, these would actually be working with file descriptors and wind up calling `fs.futimes`, `fs.fchmod` and `fs.fchown` internally. I don't see the problem with extending the adapter beyond `src` and `dest` anyway, these extensions being `fs` specific after all, but note that if this were a separate module we'd have to `open` the file a second time.

I'd also like to see a `vfs.stat` for getting these attributes from disk into the vinyl objects, but I seem to be alone in that :-)

piranna:
> Well, if we want to stick with Node fs nomenclature then we'd wind up with vfs.utimes, vfs.chmod and vfs.chown

Fair for me.

> but note that if this were a separate module we'd have to open the file a second time

Interesting... I think this is a good point about having that functionality integrated on `dest()`, but would need to be check first.


> I'd also like to see a vfs.stat for getting these attributes from disk into the vinyl objects, but I seem to be alone in that :-)

Yeah, because it's already done by `src()` :-D

erikkemperman:
> Yeah, because it's already done by src() :-D

Fair enough, I guess I had that coming -- of course I am again talking about the case where `dest` is not the last stage in a pipeline :-)

But maybe the consensus is actually that if these attributes like times etc matter especially then you should just make a distinct pipeline starting with another `src`.

piranna:
> But maybe the consensus is actually that if these attributes like times etc matter especially then you should just make a distinct pipeline starting with another src.

It makes sense, and it would be simpler...

contra:
Here are the viable options as I see them:

- `dest` always syncs all attributes to the new file (huge performance loss)
- `dest` syncs only when an option is true
- `dest` becomes intelligent and knows which attributes need syncing, performing the minimum number of fs calls required

I don't think exposing new functions for different attributes is good API design on this one.

erikkemperman:
Maybe this is a completely inaccurate projection of long-ago C intuition, but once you have a file descriptor to an open file, which you can safely re-use within `dest`, subsequent `fxxx` calls should not be all that bad in terms of performance?

The hit would have to be pretty horrible for users to accept an awkward default behaviour of `dest`, it seems to me.

contra:
@erikkemperman Only way to know for sure is to make some benchmarks

piranna:
> dest syncs only when an option is true

+1

> dest becomes intelligent and knows which attributes need syncing, performing the minimum number of fs calls required

+1000

phated:
I'm fine with all of @contra's proposals.  However, changing this now that there was a release (due to being an added feature) is a breaking change, so we need it so we need it solved ASAP.  This seems to be the last blocking thing on gulp 4

erikkemperman:
Looking at this yesterday I noticed that there is already a `chmod` in there.  But that is done in `writeContents/index.js`, while the recent addition of `(f)utimes` is one level down in `writeBuffer.js` and `writeStream.js`.

So I think it would be cleaner to handle these attributes in one place, but passing file descriptors around is a bit messy because that doesn't work for `writeDir.js`.

I also noticed that the logic around `chmod` makes a point of doing `stat` first and checking if the mode is already correct. So up to two `fs` calls in stead of just writing it unconditionally.

I don't see it yet but assume there's a good reason for that. There's even a test to check that `chmod` doesn't get called if not necessary.

piranna:
> Looking at this yesterday I noticed that there is already a chmod in there. But that is done in writeContents/index.js, while the recent addition of (f)utimes is one level down in writeBuffer.js and writeStream.js.

I absolutely missed this, sorry. If I would have been aware of this, definitely I would have done it there, and in fact I think we should move futimes there. This has the advantage that we are already calling `fs.stat()`, so there's no need to do it twice. Problem would be with directories, since there's no currently a way to get it's file descriptor (they are working on it), but we could pass to `written()` the file descriptor as second parameter and use the writePath if it's not defined, like the case for the directories.

I've also seen that by using `fs.stat()` it needs to check symlinks to ignore them since they are followed, while we should be using `fs.lstat()` to update the symlink itself.

If you want I can be able to do this changes on this pull-request or in a new one. Only thing that to do them in a clean way instead of nested callbacks and constantly checking outputs we should use `async` module to be able to call chmod, chown and futimes in parallel...

erikkemperman:
@piranna That's exactly what I was thinking; pass the `fd` back into the `written` callback, and close the open `fd`, if any, in either `written` or maybe in `complete`.

It gets slightly messy, imho, to have two blocks for setting attributes, one for when we have an `fd` and the other for when we have only `writePath`. On the other hand I suppose it would eventually get simpler again, since you mention that descriptors will be supported for directories as well (any time estimate?).

Minor annoyance there is that `writeDir` currently sometimes returns a second argument (see `mkdirp`) and that `writeSymbolicLink` always returns a second argument (though it looks like it is never used?). But that can be changed, or checked against in `written`.

Also agree to do the atime/mtime near the same place as `chmod`. That would also mean we don't really need a separate script for it (because we won't be using it from different places any longer). Or if this gets too big maybe a more general `metadata` script.

But first I guess we'll need to resolve which of @contra's alternatives would work best. I have been thinking about what would be "intelligent" in this context but can't seem to come up with anything that wouldn't be really ugly. Also, if it gets too complicated in terms of cases it will be hell to document. But maybe that's just me, any of you have ideas?

Since `chmod` and `futimes` are currently already in there, I might favour going with option 1, which was that the default would be to (try to) write the attributes. Not sure if `chown` should be in there too? We could then have options to skip this step, either generally or per attribute, or maybe an option to attempt it but suppress any `EPERM` errors.

I will run some benchmarks to confirm, but I expect the performance hit to be acceptable when we re-use descriptors as much as possible.

I am also planning to look into the exact permissions required for each of these attributes. With any luck that will suggest the natural order in which to attempt them.

Won't have any time for this today or tomorrow though.

piranna:
> since you mention that descriptors will be supported for directories as well (any time estimate?)

Don't know since it needs some support from libuv, but you can read more about this at https://github.com/nodejs/node/issues/583 and https://github.com/libuv/libuv/pull/416. As you can see, this has been discussed since almost a year ago and it's still alive...


> That would also mean we don't really need a separate script for it (because we won't be using it from different places any longer). Or if this gets too big maybe a more general metadata script.

:+1:


> But first I guess we'll need to resolve which of @contra's alternatives would work best. I have been thinking about what would be "intelligent" in this context but can't seem to come up with anything that wouldn't be really ugly. Also, if it gets too complicated in terms of cases it will be hell to document. But maybe that's just me, any of you have ideas?

I think `async` could do the trick by splitting the pieces. I understand as intelligent as trying to set as more attributes as possible and if not update the vinyl object (we have do the `stat` anyway, so this is free...).


> Since chmod and futimes are currently already in there, I might favour going with option 1, which was that the default would be to (try to) write the attributes. Not sure if chown should be in there too? We could then have options to skip this step, either generally or per attribute, or maybe an option to attempt it but suppress any EPERM errors.

+1 to all.

erikkemperman:
>  I understand as intelligent as trying to set as more attributes as possible and if not update the vinyl object (we have do the stat anyway, so this is free...)

Interesting. That's not how I interpreted "intelligent" here. Also, I was under the impression that updating the vinyl object from disk in stead of vice versa, when writing fails, was such a crazy idea on my part that it didn't even get any response earlier :-) But absolutely, since the `stat` is there already this would be quite easy to do.

I'm not a big fan of `async` but it looks like it is already in node_modules anyway (courtesy of `istanbul`). Anyway I don't think `parallel` is what you meant? That would go wrong quite spectacularly I imagine, for a single `fd`.

piranna:
> was such a crazy idea on my part that it didn't even get any response earlier

And it is :-) But the only other razonable option would be to fail due to lack of permissions, so if we add a flag to enable the assignement of the more attributes as possible, instead of simply swallow them we could be able to do something useful with them like update the vinyl object :-) I think this has the best of both worlds (pun intended :-P).


> I'm not a big fan of async but it looks like it is already in node_modules anyway (courtesy of istanbul). Anyway I don't think parallel is what you meant? That would go wrong quite spectacularly I imagine, for a single fd.

Yes, I'm only talking about `parallel()` function. I don't think it would be a problem since they are not really executed in parallel, and if so, we could use instead `serial()`. The point here it's to don't have a lot of nested callbacks and checks.

phated:
@piranna I don't think `async` is needed.  If a flow control library is absolutely needed, we already have `now-and-later` in the dep tree of gulp.  @erikkemperman istanbul isn't installed in most normal scenarios so `async` won't be in the dep tree.

The correct way to make this intelligent is to keep the file attributes in sync, including (but not limited to) making the `contents` property a getter/setter that updates the modified time, checking if a file exists before creating it with `dest`, updating the vinyl object to match the filesystem before `dest` re-emits the file, etc.

I think the above is probably the best thing to do, but I don't have the bandwidth to contribute it currently, would either of you be able to?

erikkemperman:
> making the contents property a getter/setter that updates the modified time

@phated I may be misunderstanding you there... To my mind we were discussing here the best procedure to handle whatever metadata the vinyl object happens to be carrying once it gets into `dest`. But if I understand your suggestion that's more about how that metadata (access/mod times in particular) should be updated throughout a pipeline, as plugins read/write `file.contents`?

For one thing, that would have to be done in the `vinyl` repository, not here. And actually, it looks like `File.prototype` already defines its `contents` property with a getter/setter descriptor. So to update the `stat.mtime` (set) and/or `stat.atime` (get) would be easy to add, if that is in fact the desired behaviour there.

Not to throw work in his lap :-) but it sounded to me like @piranna was getting ready to update this PR according to the latest discussion here? If not I might find some time this weekend, but can't promise anything unfortunately.

piranna:
> @piranna I don't think async is needed. If a flow control library is absolutely needed, we already have now-and-later in the dep tree of gulp

It's not absolutely needed, it's only for code cleaness, that maybe would also be achieved by a simple local function. `async` or any other flow control library are not mandatory :-)


> The correct way to make this intelligent is to keep the file attributes in sync, including (but not limited to) making the contents property a getter/setter that updates the modified time, checking if a file exists before creating it with dest, updating the vinyl object to match the filesystem before dest re-emits the file, etc.

:+1: 


> it sounded to me like @piranna was getting ready to update this PR according to the latest discussion here?

I felt guilty because these problems has surfaced due to my inclussion of utimes, so yes, I volunteer about updating this :-) More things to add to my bachelor thesis :-P

erikkemperman:
>  felt guilty because these problems has surfaced due to my inclussion of utimes

Likewise, for having jumped in with `futimes` :-)

piranna:
> Likewise, for having jumped in with futimes :-)

Then please help the [stat vs. lstat problem](https://github.com/gulpjs/vinyl-fs/blob/50c3683d389de338b9a144b5045820c851317308/lib/dest/writeContents/index.js#L45-L49), by changing it I don't know why the tests fails since they are not using symlinks... :-(

piranna:
I've find that the symlinks don't have file descriptor based functions, so although they will be implemented for directories, they will still needed be needed a path-based flavour... :-/

erikkemperman:
So currently the implementation does nothing with these attributes if it's a symlink. It might be safest, at least for now, to keep it that way. Or... What do you think the behaviour ought to be? Change metadata on the link? Or on the target?

piranna:
> What do you think the behaviour ought to be? Change metadata on the link? Or on the target?

Obviously change the metadata of the link. The problem is that by changing it I'm getting errors in unrelated tests, and seems it's neccesary to change it to be able to move the file descriptors to the `writeContents/index.js` file... :-/

erikkemperman:
It's obvious? So why is the current implementation skipping the metadata in this case?

Sorry I'm not able to spend the time this deserves right now, kind of in the middle of something at my day job...

piranna:
> It's obvious?

Yes, symlinks are directory entries with their own metadata too.


> So why is the current implementation skipping the metadata in this case?

Seems [POSIX ignore symlink permissions](http://unix.stackexchange.com/questions/224979/why-do-linux-posix-have-lchown-but-no-lchmod) and use the final file ones, but OSX does. Maybe that's the reason why `lchmod()` is doing nothing.

piranna:
Ok, found the lstat problem, it was due to the usage of spies on the tests (I'm not used to them).

erikkemperman:
Sorry, yeah, I know what symlinks are :-) It's just not obvious to me, at a glance, that we'd want to write metadata to the _link_ or to its _target_ -- given that the current implementation is to do nothing with metadata in this case.

piranna:
I have managed close the file descriptors on writeContent/index.js, but don't know why after moving there the futimes tests fails randomly, and only constant behaviour is that sometimes `test.coffee` content gets changed from "Hello world!" to "1234567890" and the next error:

```
  1) dest stream "after each" hook:
     Uncaught Error: EBADF: bad file descriptor, read
      at Error (native)
```

I've uploaded the code to see if someone is able to find why this doesn't work, the commit before the last one works correctly, but I have the feeling that I've lost the full day doing this... :-(

piranna:
Ok, I've found it! :-D Problem was due to some sort of race condition, since it only worked correctly if futimes had its own call to fs.stat(). Changing it for process.nextTick() worked a little better, but the "Bad File Descriptor" problem was still there, but by using old `setTimeout(f, 0)` works flawlessly :-) I have tested it several times without failure, so I'll clean up it a little bit and add the changes I did before and I'll update this pull-request.

piranna:
I've finished the clean-up and updated the pull-request, now it should work correctly and work as basis to see how to improve this in the future.

piranna:
The tests are failing on Node.js 0.10 because I've hardcoded the file descriptor to mimic the "expectedPath" parameter used before (seems it's somewhat constant), if you have a better idea about how to check this or maybe also if that check could be removed please tell me.

erikkemperman:
@piranna Sorry I haven't been able to look at this over the weekend -- a festival sort of just happened. Today I'll be swamped in some things at work but hopefully I'll have some energy left to take this up again this evening (CET).

In particular I'm curious why the `setTimeout(..., 0)` is needed. Also I was a bit worried that maybe the variable `outFD` I introduced in `writeStream` might get mixed up in case of concurrency... Did you notice anything like that?

erikkemperman:
@phated Also, still wondering what you were getting at when you mentioned getter/setter accessors on `file.contents`... Care to elaborate a bit?

piranna:
> In particular I'm curious why the setTimeout(..., 0) is needed

I'm not sure, but as I told before, without it I get a lot of random error, all of them started by a "Bad file descriptor" error. Using `process.setInmediate()` fix them a bit but they still happen, but by using `setTimeout()` or `fs.stat()` don't. Seems to be a race condition or something.


> Also I was a bit worried that maybe the variable outFD I introduced in writeStream might get mixed up in case of concurrency... Did you notice anything like that?

I'm not sure, but I think it's the way to go if we want reduce the access to the filesystem, and it's necesary to have sub-second precission with the usage of `futimes()`. Other option would be to revert all changes and keep using only filepath based functions, in any case the files will be aready on the cache so time access would be despreciable and the could would be simpler...

erikkemperman:
All right, like I said, I hope to be able to keep my eyes open long enough to look at this tonight.

piranna:
> All right, like I said, I hope to be able to keep my eyes open long enough to look at this tonight.

Hahaha :-D Ok ;-)

erikkemperman:
@piranna Just fetched your commits but unfortunately some tests are failing here (MacOSX, Node 4.1.2). Can't see straight right now, long day after a short night. Will look again tomorrow, let you know what I find.

piranna:
Sleep, you have earned it :-)

2015-11-23 19:03 GMT+01:00 Erik Kemperman <notifications@github.com>:

> @piranna <https://github.com/piranna> Just fetched your commits but
> unfortunately some tests are failing here (MacOSX, Node 4.1.2). Can't see
> straight right now, long day after a short night. Will look again tomorrow,
> let you know what I find.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gulpjs/vinyl-fs/pull/128#issuecomment-159013499>.
>



-- 
"Si quieres viajar alrededor del mundo y ser invitado a hablar en un monton
de sitios diferentes, simplemente escribe un sistema operativo Unix."
– Linus Tordvals, creador del sistema operativo Linux

phated:
@erikkemperman sorry, I thought the conversation had diverged to much to follow up on that question.  I'll elaborate.

Currently, vfs is stating every file to figure out if it is a symlink and the attaches those stats to the vinyl object.  If you do something like the following, the modified time of each vinyl object will still be the original time as stat'd from the fs.
```js
vfs.src('**/*.js')
  .pipe(modifyContentsPlugin()) // assume some plugin that modifies contents
  .pipe(vfs.dest('./dist/');
```

My point was that we don't actually maintain proper file state in the vinyl objects in all scenarios, which will make the intelligent updates hard without many underlying changes.  However, I think that is the correct path forward, but I just wanted to outline some of the problems I saw.

Does that help?

erikkemperman:
>  I thought the conversation had diverged to much to follow up on that question. I'll elaborate.

It has, no doubt :-) Thanks for clearing up.

If maintaining proper file state is the way forward, changing `stat.atime` and `.mtime` from the getter and setter, respectively, of `file.contents` might help. But as those accessors already exist it seemed deliberate that this isn't already happening?

phated:
@erikkemperman it was not deliberate, it is just the simplest implementation.  No consideration was ever made for `file.stat` because we didn't read them in before the symlink functionality was added.

erikkemperman:
@piranna
So, I spent some time with your changes. One problem is that we can't expect the `expectedFD` to equal 12 -- on my Mac it is mostly 17, but not always. So I suggest something like [this](https://github.com/erikkemperman/vinyl-fs/commit/08c70351a1dfbdac324d0aa8139b0b675d245ff4).

There are intermittent and varying problems with the before/after hooks (`wipeOut`) in the `dest` tests. The `setTimeout(..., 0)` you added makes these go away... But unfortunately not always! For instance, Travis now failed for node 0.12, see [here](https://travis-ci.org/erikkemperman/vinyl-fs/jobs/92989819). I think the node version might not be relevant, actually.

The fact that `setTimeout` makes these errors happen less often might point us toward the underlying problem here, but until we're sure what's going wrong it's probably unhelpful to have it?

I don't really understand this `wipeOut` implementation, it calls `rimraf` with a callback and then resets some spies. Seems to me those resets should happen before `rimraf` or inside its callback?

Unfortunately this is all the time I have for this now :-(

piranna:
> So, I spent some time with your changes. One problem is that we can't expect the expectedFD to equal 12 -- on my Mac it is mostly 17, but not always. So I suggest something like this.

LGTM. @phated?


> The setTimeout(..., 0) you added makes these go away... But unfortunately not always!

Oh, sh*t... I should have suspected it when `process.setInmediate()` didn't work :-(


> I don't really understand this wipeOut implementation, it calls rimraf with a callback and then resets some spies. Seems to me those resets should happen before rimraf or inside its callback?

Don't know, I didn't do them.


> Unfortunately this is all the time I have for this now :-(

Fair enough, good job :-)

phated:
Why is the expectedFd not always 12?

piranna:
> Why is the expectedFd not always 12?

Probably is system dependent, maybe the order is different or the fd are not reused. I admit my solution was not a clean one, but was not able to think any better.

phated:
I don't think that the number of FDs should be system dependent, but I'm not sure.

erikkemperman:
I'm not sure either, but not totally surprised that the value of fd here might differ per platform. Or per node version (internally used fds come from the same pool I think).

For what it's worth, I believe the second part of the `if` (that I changed in the commit I linked above) might be superfluous anyway... But maybe I'm missing something obvious?

Either way, it seems to me that the problem in before/after in `test/dest.js` is the main hangup here. But I haven't been able to figure out much about that in the little time I had.

Really busy at the university where I work, so I might not be able to invest much more time either...

Does the implementation of `wipeOut` look dodgy to you as well?

ksnyde:
https://github.com/stevelacy/gulp-bump/issues/47#issuecomment-161398118

ksnyde:
I've just scanned through this issue and it appears to be the same issue I ran into today. Hopefully this example will help (although it may fit into the "yeah, already got it" category).

https://github.com/stevelacy/gulp-bump/issues/47

erikkemperman:
@ksnyde I think this issue here is only half of the problem -- writing metadata like modified stamps to disk from `dest`. The other half would be an accurate accounting of this metadata in the vinyl objects throughout the pipeline.

An earlier attempt at writing time stamps from `dest` was already merged, so if we can get the second half fixed (in the `vinyl` repo) you might be good to go.

The reason we were looking at improving upon what was merged earlier (here in `vinyl-fs`) had to do with ownership, which it sounds like wasn't a problem in your case?

lib/dest/writeContents/index.js
```diff
@@ -36,27 +100,65 @@ function writeContents(writePath, file, cb) {
...
+    if (isErrorFatal(err)) {
+      return complete(err);
+    }
+
+    if (!file.stat) {
+      return complete();
+    }
+
+    if (typeof fd === 'number') {
+      return fs.fstat(fd, applyStatFd);
```
phated: why is there separate paths for `fd` and non-`fd`?

piranna: To use the fd-based functions when they are available. This improve
performance by preventing some access to the filesystem, and also they are
needed to have sub-second precision. The path based ones are still needed
though, because fd ones only can be applied to files, not directories.

phated: @piranna thanks for the quick response, I really want to wrap up the vinyl-fs issues/PRs so we can get gulp 4 shipped so I'll be taking some time to look over this thoroughly

piranna: You are welcome :-) Hope we can find a solution to this issue soon.

phated: @piranna I just attempted to use `fs.open()` on a directory with mode `'r'` and it returned a `fd` to the directory, are you sure we can't use `fd`s everywhere?

phated: @piranna just tested `fs.fchmod` on a read-only file descriptor to a directory and it works.  Could you simplify this PR by only using file descriptors?

piranna: Wow, that's surprising... Maybe the point is that you are fetching a file
descriptor for the directory and not trying to use some fancy functions
like opendir() (that's not available on Node.js yet...)? What version of
Node.js are you using? I'm afraid it would only works on the latests ones
or we are using here an implementation bug, but if there are no more
concerns, I'm +1 to use that hack if it's working for our purposses :-)

phated: I'm using node 0.10 (the furthest back we support).  I did some googling and it seems that it is expected behavior as long as the flag is set as read-only instead of read-write

phated:
@piranna @erikkemperman would it be beneficial to use https://www.npmjs.com/package/fd as a dep to manage the file descriptors?

piranna:
> @piranna @erikkemperman would it be beneficial to use https://www.npmjs.com/package/fd as a dep to manage the file descriptors?

I'm not sure... :-/

I forgot about other point: both creation of symlinks and directories
don't return a file descriptor, that's the reason why we were having
the path-based functions too. We could do the trick for directories by
opening it after it's created if we want to set those values, but what
about symlinks? If we try to do it, the fd will be from the symlink or
from the file it's pointing?


-- 
"Si quieres viajar alrededor del mundo y ser invitado a hablar en un
monton de sitios diferentes, simplemente escribe un sistema operativo
Unix."
– Linus Tordvals, creador del sistema operativo Linux

erikkemperman:
rvagg's lib looks like it might help sorting out that "Bad descriptor" issue without actually figuring it out:
> fd manages fs.open() and fs.close() calls safely for you where there may be timing issues related to multiple-use of the same descriptor.

I'm still curious why the first implementation of symlinks skipped the `chmod` entirely, when everything was done with `writePath`. But then I hadn't expected `dest` to do symlinks at all, actually. Not really a thing you do with streams, I supposed.

Would it make sense to keep skipping metadata for symlinks?

piranna:
> But then I hadn't expected dest to do symlinks at all, actually. Not really a thing you do with streams, I supposed.

cpio-stream send the destination of symlinks as a stream, and it makes sense since it's its content :-)


> Would it make sense to keep skipping metadata for symlinks?

Maybe, since seems Posix ignore them at all and only pay attention to the stats of the destination file/folder. I think it's a bug in the Posix spec, but what the h*ll...

piranna:
I've just updated the pull-request to unify the branches for fd and path methods, but I've found a problem in one of the tests where is being created a symlink to a non-existent file. Now by using `.open()` to get the file descriptor (because they are not being generated for directories nor symlinks) it try to open the file instead of the symlink, so it fails. It's possible to make `.open()` to open the symlink itself, but it requires to do so using the flags `O_PATH` and `O_NOFOLLOW`, that are not being used by Node.js itself. Here we have three options:

* made mandatory that all symlinks point to an existent file, so the file operations works on that destination (Posix seems that's the way to do it)
* use `fs.lstat()` as it was being done before for this use case
* ignore the `.open()` error and the file operations if the destination don't exists

lib/dest/writeContents/writeStream.js
```diff
@@ -3,14 +3,14 @@
...
-var futimes    = require('../../futimes');
 
 function writeStream(writePath, file, cb) {
   var stat = file.stat;
 
   var opt = {
     mode: stat.mode,
-    flag: file.flag
+    flag: file.flag,
+    autoClose: false
```
phated: is this used for something?

lib/dest/writeContents/writeStream.js
```diff
@@ -43,7 +37,7 @@ function writeStream(writePath, file, cb) {
     outStream.removeListener('error', complete);
     outStream.removeListener('open', open);
     outStream.removeListener('finish', success);
-    cb(err);
+    cb(err, outFD);
```
phated: where is outFD defined?

phated:
Other than 1 failing test and the question I posted on the diff, I am feeling pretty good about these changes.  I need some more eyes on the entire diff though, @contra @erikkemperman - do either you have time to do an in-depth review?

lib/dest/writeContents/writeStream.js
piranna: On the `open` event.

contra:
@phated Can you provide a summary of the changes involved in all of these PRs/issues that have been going on for the last week? I can review it a lot faster that way.

lib/dest/writeContents/index.js
```diff
@@ -36,26 +90,55 @@ function writeContents(writePath, file, cb) {
...
       return complete();
     }
 
-    fs.stat(writePath, function(err, st) {
+    if (typeof fd === 'number') {
+      return stat(fd);
+    }
+
+    fs.open(writePath, 'r', function(err, fd)
+    {
```
contra: style doesnt fit our guidelines

lib/dest/writeContents/index.js
```diff
@@ -36,26 +90,55 @@ function writeContents(writePath, file, cb) {
...
     }
 
-    if (!file.stat || typeof file.stat.mode !== 'number' || file.symlink) {
+    if (!file.stat) {
       return complete();
     }
 
-    fs.stat(writePath, function(err, st) {
+    if (typeof fd === 'number') {
+      return stat(fd);
```
contra: complete never gets called?

piranna:
> From a quick glance the control flow needs quite a bit of cleanup, the code is a little messier than usual.

That's because we have a lot of optional behaviours here: chmod, chown, futimes... That's why it was propossed before to use `async.parallel`, to make each one independent of the others. If vinly-fs don't use any flow control module, maybe we can craft an ad-hoc function here...

lib/dest/writeContents/index.js
piranna: Fixed. Could this be checked on a `pre-commit` hook?

contra:
@piranna I'm okay with whatever module you want to use, personally I use async across projects so that would be my recommendation. I think there was an objection over bringing in another module but I care about code quality and not shaving dependencies and bytes off server-side modules.

piranna:
I use `async` by default in a lot of projects, if there's another alternative it's welcome. I'll use what @phated will say :-)

erikkemperman:
@phated @contra @piranna
I think `now-and-later` was suggested for flow control earlier, this is already in the dep tree through `undertaker`.

Tomorrow I'll have some time to look into this PR again, but for now I notice that

- the expected fd in the tests is still `12` which will fail on Mac for some reason. Something like [this](https://github.com/erikkemperman/vinyl-fs/commit/08c70351a1dfbdac324d0aa8139b0b675d245ff4) would work
- the `setTimeout(...,0)` is still in there, seems to me this is suppressing a symptom of a deeper problem. Maybe rvagg's `fd` library would help.

erikkemperman:
Sorry guys, looks like I never actually posted what I found... Which is little more than I mentioned above unfortunately. I've been really busy at work the last couple of weeks but now I have a few days off and hopefully I'll be able to finally sort some of this out.

Ah, and happy holidays!

phated:
@piranna there were some changes that got merged and one fixed a problem with streaming contents.  Can you rebase this PR on top of the latest changes?

piranna:
Lol, I was just reading the topic :-P I'll rebase the changes :-)

phated:
@piranna will you have a chance to rebase this soon?

contra:
Going to put a time limit of a few more days on these stat changes - if we aren't able to get the fixes merged soon I'm going to revert the broken changes that were merged earlier since it is holding up a release.

phated:
@contra while I agree that it is annoyingly holding up a release, I don't know if removing the functionality to push a release out quicker is the best idea.  It would require bumping vinyl-fs a major version since we are removing functionality that was added in a minor and then I would only feel comfortable adding it back in a major since it caused all these troubles, which means it will never land in gulp 4.  This feature surfaced a big problem with vinyl and vinyl-fs - the fact that they only deal with a tiny subset of file properties (path and contents) but it is advertised as an in-memory, streaming file system abstraction.  If we aren't handling other metadata correctly and aren't able to update that metadata on the actual filesystem, people are going to reach for the `fs` module in their plugins, which goes against one of the core tenants.

contra:
@phated Was this ever released on npm in a way that people could use?

phated:
@contra yeah, unfortunately the utimes stuff was released in 2.2.0.  2.2.1 was a fix for trying to write invalid dates, 2.3.0 now has a fix for sub-second precision and these changes are to fix a bug where the utimes call fails if the process user isn't the file owner.  We also need to finish https://github.com/gulpjs/vinyl/issues/72 (which @erikkemperman started on) because we actually read the stats of every file and attach them while looking for symlinks (we didn't do that before 2.x)

We found all these issues because people are using it with the gulp 4.0 branch and have been reporting the problems.

piranna:
> @piranna will you have a chance to rebase this soon?

Sorry for the delay, I have been busy lately and the rebase is giving me a lot of problems to make the tests to pass.


> This feature surfaced a big problem with vinyl and vinyl-fs - the fact that they only deal with a tiny subset of file properties (path and contents) but it is advertised as an in-memory, streaming file system abstraction. If we aren't handling other metadata correctly and aren't able to update that metadata on the actual filesystem, people are going to reach for the fs module in their plugins, which goes against one of the core tenants.

It seems the module [fstream](https://github.com/npm/fstream) from the npm developers is getting some traction (I didn't knew of it until some weeks ago and now I'm seeing it's a dependency of almost all the projects I'm using lately...) and has support for streaming directories and symlinks,  maybe we could get some inspiration from it...

piranna:
I can't be able to fully fix the problems that has surfaced after the rebase, seems it's getting some problems due to the usage of the file descriptors... Don't know if the library to manage them would help, but the more I think about it, the more I believe we've gone too far optimizing by using them instead of using plain old path based operations just to have sub-second precission... Does it matter so much?

Also, other problems are related with symlinks and the fact they are not used as first class citizens here... :-/

Klowner:
Sorry I'm a little late to the show, but what issue are the symlinks posing? I might be of some assistance.

erikkemperman:
I attempted the rebase of this stuff as well, and couldn't figure it out either. Didn't spend a lot of time on it though, but my feeling from its description was this library from rvagg might help.

Personally I feel that sub-second precision is rather a big deal. That is, if we want things like `gulp-newer` and `gulp-changed` to work as expected.

piranna:
> Sorry I'm a little late to the show, but what issue are the symlinks
posing? I might be of some assistance.

As I remember, since some filesystems don't support symlinks (mostly FAT)
to be as compatible as possible vinyl-fs dereferenced the symlinks writting
instead its content, so it was being used fs.stat() instead of fs.lstat()
and didn't take on consideration the stat of the symlinks. It was
succesfully changed, but when we started using file descriptors we used
fs.fstat(), and that one show the info of the final file, not the symlink.

erikkemperman:
@contra @phated @piranna @Klowner

So I spent a little more time with the rebased work and got all tests to pass. Yay!

See the code [here](https://github.com/erikkemperman/vinyl-fs/tree/feature-metadata) and the test results [here](https://travis-ci.org/erikkemperman/vinyl-fs/builds/102135663).

I've taken the liberty of squashing the very many commits, I hope you don't mind...

I guess it should still be tidied up a bit with some flow-control library, like we talked about earlier.

More importantly, it currently skips writing metadata writing for symlinks, just like it did before... I propose discussing that in a different issue if anyone has ideas.

Klowner:
nice job, @erikkeperman! That seems good to me from what I can tell.

erikkemperman:
> All this commits will be lost as tears in the rain... :-P

How appropriate, it was his [birthday](http://www.theregister.co.uk/2016/01/09/roy_battys_birthday/) last Saturday :+1: 

Seriously though, we can obviously package this however you like -- to use a football phrase, I just walked into a header from an excellent pass.

To reflect this you might consider adding my fork as a remote, and force-pushing it into your branch for this PR? Feel free to squash as you see fit :-)

Like I said it probably still needs some work, for one thing I really don't like that descriptors and streams are opened and closed/ended in different modules now. Maybe the `written` callback in `writeContents/index` should be given a back-callback (is there a name for such a thing?) argument, invoke after it is done writing metadata. The various `writeContents/writeXxx` modules can then close the descriptor or end the stream, as appropriate

Come to think of it, that might even make things sufficiently separated to leave out further flow-control.

erikkemperman:
I think I made it slightly more legible... But that's subjective of course.

Will leave it for now, let me know how you want to proceed!

phated:
@erikkemperman you are my hero today.  Can you squash your commits so there's just 2 commits (1 squashed @piranna commit and 1 squashed commit for you) and then submit a PR?

piranna:
> All this commits will be lost as tears in the rain... :-P
> 
> How appropriate, it was his birthday last Friday :+1:

I was not thinking about that fact, just I love that movie :-D


> one thing I really don't like that descriptors and streams are opened and closed/ended in different modules now

That's because close of file descriptors is common, probably the same could be done with open.

phated:
Completed by #151 - released in 2.3.2

Thanks for all the work you did on this @piranna!

piranna:
You are welcome :-)

2016-02-26 2:19 GMT+01:00 Blaine Bublitz <notifications@github.com>:

> Completed by #151 <https://github.com/gulpjs/vinyl-fs/pull/151> -
> released in 2.3.2
>
> Thanks for all the work you did on this @piranna
> <https://github.com/piranna>!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gulpjs/vinyl-fs/pull/128#issuecomment-189067162>.
>



-- 
"Si quieres viajar alrededor del mundo y ser invitado a hablar en un monton
de sitios diferentes, simplemente escribe un sistema operativo Unix."
– Linus Tordvals, creador del sistema operativo Linux

Issue #129, opened 2015-11-12
base option possibly not reflecting correct default value


treshugart:
Forgive me if I'm not understanding something correctly here. In the `base` option, you say:

> Default is where the glob begins if any.

This I *think* I understand correctly in that if you specify `something/**` that `base` will be `something` and `**` will be the glob.

However, you also say:

> Default is `process.cwd()` if there is no glob.

What is considered "not a glob"? Is this something that doesn't contain a pattern? An empty string? A relative path? An absolute path?

If an absolute path classifies as a non-glob argument, then if you do something like:

```js
vinylFs.src(['/file/path1.js', '/file/path2.js']);
```

Then the `base` is automatically set as `/file` instead of whatever the current working directory is. That said, I can kind of see the point of this because if you had two completely separate globs that the base path would be ambiguous. But then why have it default at all? The only way a base path could be unambiguous would be if you passed a string glob instead of an array.

In order to get around this you can force the `base` option via the second argument but I'm still wondering if this is the correct behaviour.

phated:
You are specifying absolute paths.  All paths are resolved using node's `path.resolve` method, so if we do `path.resolve(process.cwd(), '/file/path1.js')` it is going to result in the absolute path of `/file/path1.js`.  To make your paths relative, prefix them with `.` or don't put the `/` as is standard unix convention.

treshugart:
I understand the unix convention and node's path resolution semantics but that wasn't my question. My question is: what is considered "not a glob"?

And depending on that answer, why is the `base` reporting as the `dirname` for the files? Shouldn't it be `process.cwd()` like the docs say?

phated:
Anything that doesn't have glob magic is "not a glob".  The `base` property on vinyl objects is resolved by glob2base at https://github.com/gulpjs/glob-stream/blob/master/index.js#L27 but will be switching to glob-parent in https://github.com/gulpjs/glob-stream/pull/56 which has the same behavior.

treshugart:
> Anything that doesn't have glob magic is "not a glob".

So:

> Default is process.cwd() if there is no glob.

Means that `base` should be `process.cwd()` instead of `/file` for `/file/path.js` right?

phated:
No, it gets resolved

PR #130, opened 2015-11-12
#129 - Update documentation to more clearly explain what happens for default base option values.

treshugart:
After discussing it with @phated in #129, it seems to me that the previous description of the default `base` value is incorrect in the docs. I was originally unsure where the source of truth was (docs or code) but it seems that the code is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ fs.src(['*.js', '!b*.js'])
     - Default is `process.cwd()`.
 
   - base - Specify the folder relative to the cwd. This is used to determine the file names when saving in `.dest()`.
-    - Default is where the glob begins if any.
-    - Default is `process.cwd()` if there is no glob.
+    - Default is where the glob begins, if any. For example, `path/to/**/*.js` would resolve to `path/to`.
+    - If there is no glob (i.e. a file path with no pattern), then the dirname of the path is used. For example, `path/to/some/file.js` would resolve to `path/to/some`.
 
   - buffer - `true` or `false` if you want to buffer the file.
     - Default value is `true`.


README.md
```diff
@@ -42,8 +42,8 @@ fs.src(['*.js', '!b*.js'])
     - Default is `process.cwd()`.
 
   - base - Specify the folder relative to the cwd. This is used to determine the file names when saving in `.dest()`.
-    - Default is where the glob begins if any.
-    - Default is `process.cwd()` if there is no glob.
+    - Default is where the glob begins, if any. For example, `path/to/**/*.js` would resolve to `path/to`.
+    - If there is no glob (i.e. a file path with no pattern), then the dirname of the path is used. For example, `path/to/some/file.js` would resolve to `path/to/some`.
```
treshugart: This is sort of an implementation detail in that `glob2base` is used (and that glob-parent may be used in the future), however this hopefully better describes what happens even if it doesn't describe exactly what `glob2base` is doing internally.

README.md
```diff
@@ -42,8 +42,8 @@ fs.src(['*.js', '!b*.js'])
     - Default is `process.cwd()`.
 
   - base - Specify the folder relative to the cwd. This is used to determine the file names when saving in `.dest()`.
-    - Default is where the glob begins if any.
-    - Default is `process.cwd()` if there is no glob.
+    - Default is where the glob begins, if any. For example, `path/to/**/*.js` would resolve to `path/to`.
```
treshugart: Thought I'd add an example, especially since the second point to this behaves differently and also has an example.

phated:
Thanks!

PR #131, opened 2015-11-29
Fixes #132 The symlink following logic in src() was incorrect resulting in wrong file.base being generated.

### LINKED ISSUE #132: Symlink following logic in src() seems wrong to me
Basically, the generated vinyl file base should be what's set to to the symlink itself:

Say the src directory structure is like:
```
|-- foo
    |-- bar -> baz
    |-- baz
        |-- qux.txt
```
the symlink should produce: ./target/foo/bar at the end with qux.txt in it if dest is set to target. In practice it doesn't.

The reason is the file base should always be set to ./foo for anything linked by bar whereas the original recursive logic would produce foo/baz.

Such incorrect logic would yield a flattened file structure and empty directories at the end:

```
pipe(gulp.dest('target'))

the original logic would generate:

|-- target
    |-- foo
    |-- baz (empty)
    |-- qux.txt

whereas the fixed logic generates:

|-- target
    |-- foo
        |-- bar
            |-- qux.txt
```

I've opened a PR here: https://github.com/gulpjs/vinyl-fs/pull/131
Please check it out.

xunnanxu:
Basically, the generated vinyl file base should be what's set to to the symlink itself:

foo/bar -> baz/qux.txt
this should produce: foo/baz/qux.txt

but the file base should still default to ./foo whereas the original recursive logic would produce foo/baz.

Such incorrect logic would yield a flattened file structure and empty directories at the end.

### DIFF
--- a/lib/src/wrapWithVinylFile.js
+++ b/lib/src/wrapWithVinylFile.js
@@ -2,8 +2,9 @@
 
 var through2 = require('through2');
 var fs = require('graceful-fs');
+var File = require('vinyl');
 
-function resolveSymlinks(options) {
+function wrapWithVinylFile(options) {
 
   // A stat property is exposed on file objects as a (wanted) side effect
   function resolveFile(globFile, enc, cb) {
@@ -15,7 +16,15 @@ function resolveSymlinks(options) {
       globFile.stat = stat;
 
       if (!stat.isSymbolicLink() || !options.followSymlinks) {
-        return cb(null, globFile);
+        var vinylFile = new File(globFile);
+        if (globFile.symlinkTarget) {
+          // If we reach here, it means there is at least one
+          // symlink on the path and we need to rewrite the path
+          // to its original value.
+          // Updated file stats will tell getContents() to actually read it.
+          vinylFile.path = globFile.symlinkTarget;
+        }
+        return cb(null, vinylFile);
       }
       if (!globFile.symlinkTarget) {
         globFile.symlinkTarget = globFile.path;
@@ -29,7 +38,7 @@ function resolveSymlinks(options) {
         globFile.path = filePath;
 
         // Recurse to get real file stat
-        resolveFile(globFile, enc, cb);
+        return resolveFile(globFile, enc, cb);
       });
     });
   }
@@ -37,4 +46,4 @@ function resolveSymlinks(options) {
   return through2.obj(options, resolveFile);
 }
 
-module.exports = resolveSymlinks;
+module.exports = wrapWithVinylFile;
--- a/lib/src/wrapWithVinylFile.js
+++ b/lib/src/wrapWithVinylFile.js
@@ -39,7 +39,7 @@ function wrapWithVinylFile(options) {
         globFile.path = filePath;
 
         // Recurse to get real file stat
-        return resolveFile(globFile, enc, cb);
+        resolveFile(globFile, enc, cb);
       });
     });
   }
--- a/lib/src/wrapWithVinylFile.js
+++ b/lib/src/wrapWithVinylFile.js
@@ -17,17 +17,18 @@ function wrapWithVinylFile(options) {
 
       if (!stat.isSymbolicLink() || !options.followSymlinks) {
         var vinylFile = new File(globFile);
-        if (globFile.symlinkTarget) {
+        if (globFile.originalSymlinkPath) {
           // If we reach here, it means there is at least one
           // symlink on the path and we need to rewrite the path
           // to its original value.
           // Updated file stats will tell getContents() to actually read it.
-          vinylFile.path = globFile.symlinkTarget;
+          vinylFile.path = globFile.originalSymlinkPath;
         }
         return cb(null, vinylFile);
       }
-      if (!globFile.symlinkTarget) {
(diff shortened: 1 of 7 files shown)


test/src.js
```diff
@@ -385,22 +385,35 @@ describe('source stream', function() {
...
+    var expectedPath = path.join(__dirname, linkTarget);
 
-    var stream = vfs.src('./fixtures/test-symlink-dir', {cwd: __dirname});
-    stream.on('data', function(file){
+    var stream = vfs.src(linkTarget, {cwd: __dirname, base: __dirname});
+    stream.on('data', function(file) {
+      file.base.should.equal(__dirname);
+      // the path should be the symlink itself
       file.path.should.equal(expectedPath);
+      // but the stats should have be updated
```
justinsa: "but the stats should be updated", you have an extraneous "have" in there.

phated:
@xunnanxu I just had a striking realization that this might be the exact same thing as https://github.com/gulpjs/gulp/issues/1427

I just ran your branch against the case outlined in that issue and it seems to have fixed it.  Do you have a chance to rebase this onto master?

phated:
It seems that this would solve the linked issue but doesn't actually create the symlinks that exist in the original source; instead, it writes the files with the contents of the symlinked files. There is still another problem related to `{ followSymlinks: false }` but this fixes the default case of `{ followSymlinks: true }`

phated:
Is there a cleaner way to implement this?  Mutation passing across stream boundaries seems really messy to me.

GregDracoulis:
@phated what is your concern there? setting `symlinkTarget` and `path`, then using `symlinkTarget` to change `path` back?

I almost feel like the logic in `resolveSymlinks` should be worked into `createFile` instead of having to pass that mutation in. After all, it's only used that one place.

phated:
@GregDracoulis yeah, I wonder if it could be handled in a different way in vinyl-fs or maybe even in vinyl proper?  I could see resolveSymlinks possiblt being inlined, too

xunnanxu:
@phated Sorry I was out of town for a while and just got back recently. I'll try to rebase master and see if I can make it better when I have a chance

xunnanxu:
@phated I rebased master and it seems AppVeyor doesn't like the test while travis works fine and so is my local env. Can you confirm if it's configured properly? How is that different from travis and why do we need that?

xunnanxu:
@phated 
"It seems that this would solve the linked issue but doesn't actually create the symlinks that exist in the original source; instead, it writes the files with the contents of the symlinked files. There is still another problem related to { followSymlinks: false } but this fixes the default case of { followSymlinks: true }"

followSymlinks: false will work as it simply wraps the symlink itself and that behavior is already covered by tests (below the ones I added).

phated:
@xunnanxu correct but our logic attempts to traverse a symlink'ed directory and explodes.  I think it is a separate issue than this.

Appveyor is CI for windows.  Symlinks don't really work on windows, so try to get it working there but if it isn't possible, just add a skip if running on windows (see some of the chmod/utimes tests).

xunnanxu:
@phated is there an issue related to what you said? I'm not sure if I really understand what the other problem is. We may need to open another PR to address that one.

phated:
@xunnanxu no issue, just something I stumbled upon while poking around with your original issue, this PR and the linked gulp issue.  I'll create an in-depth issue (with failing test) once we get this PR sorted out.

xunnanxu:
@phated nailed. Please continue reviewing when you have a chance.

lib/src/wrapWithVinylFile.js
```diff
@@ -24,11 +35,10 @@ function resolveSymlinks(options) {
           return cb(err);
         }
 
-        globFile.base = path.dirname(filePath);
         globFile.path = filePath;
 
         // Recurse to get real file stat
-        resolveFile(globFile, enc, cb);
+        return resolveFile(globFile, enc, cb);
```
phated: no reason for return

lib/src/wrapWithVinylFile.js
```diff
@@ -16,7 +16,18 @@ function resolveSymlinks(options) {
...
+        if (globFile.symlinkTarget) {
+          // If we reach here, it means there is at least one
+          // symlink on the path and we need to rewrite the path
+          // to its original value.
+          // Updated file stats will tell getContents() to actually read it.
+          vinylFile.path = globFile.symlinkTarget;
+        }
+        return cb(null, vinylFile);
+      }
+      if (!globFile.symlinkTarget) {
```
phated: I think this should be inside the `fs.realpath` callback before we modify the `globFile.path`

phated:
@xunnanxu looking pretty good, I just had a couple of comments on things I'd like changed.

lib/src/wrapWithVinylFile.js
phated: I understand but I still think it doesn't need to happen if the realpath errors

phated:
Merged + squashed with the new github feature.  I also added a commit moving the `originalSymlinkPath` assignment inside the `realpath` callback and tests still pass.

phated:
Published as 2.4.3 - many thanks to @xunnanxu

PR #133, opened 2015-12-10
fix TypeError when the dest sourcemaps option is a boolean

chocolateboy:
This PR fixes a TypeError that's thrown when a boolean `sourcemaps` option is passed to `dest` e.g.:

```javascript
fs.dest(dir, { sourcemaps: true })
```

### test.js

```javascript
'use strict'

let coffee = require('gulp-coffee')
let fs     = require('vinyl-fs')
let rename = require('gulp-rename')

fs.src('./input.coffee', { sourcemaps: true })
    .pipe(coffee({ sourceMap: true }))
    .pipe(rename('output.js'))
    .pipe(fs.dest('.', { sourcemaps: true }))
```

### input.coffee

```coffeescript
console.log 'Hello, world!'
```

    $ node ./test.js

### output

```
test/node_modules/gulp-sourcemaps/index.js:144
    options.includeContent = true;
                           ^

TypeError: Cannot assign to read only property 'includeContent' of true
  at Object.write (test/node_modules/gulp-sourcemaps/index.js:144:28)
  at Object.dest (test/node_modules/vinyl-fs/lib/dest/index.js:28:30)
  at Object.<anonymous> (test/test.js:10:14)
  at Module._compile (module.js:425:26)
  at Object.Module._extensions..js (module.js:432:10)
  at Module.load (module.js:356:32)
  at Function.Module._load (module.js:313:12)
  at Function.Module.runMain (module.js:457:10)
  at startup (node.js:138:18)
  at node.js:974:3
```

### DIFF
--- a/lib/dest/index.js
+++ b/lib/dest/index.js
@@ -25,7 +25,12 @@ function dest(outFolder, opt) {
     return saveStream;
   }
 
-  var mapStream = sourcemaps.write(opt.sourcemaps.path, opt.sourcemaps);
+  var sourcemapOpt = opt.sourcemaps;
+  if (typeof sourcemapOpt === 'boolean') {
+    sourcemapOpt = { sourcemaps: sourcemapOpt };
+  }
+
+  var mapStream = sourcemaps.write(sourcemapOpt.path, sourcemapOpt);
   var outputStream = duplexify.obj(mapStream, saveStream);
   mapStream.pipe(saveStream);
 
--- a/test/dest.js
+++ b/test/dest.js
@@ -61,6 +61,31 @@ describe('dest stream', function() {
     }
   });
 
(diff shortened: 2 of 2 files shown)


phated:
Can you add a test for this?

chocolateboy:
Can you? :-) If there were any existing source-map tests I could have cribbed from I would have added one...

phated:
@chocolateboy it seems that the construction throws with the current implementation, can't you just write a test to make sure it doesn't throw?

chocolateboy:
@phated: [Done](https://github.com/gulpjs/vinyl-fs/commit/df874af6c3dce1b8f008230b06f3ef9ef0e30d1a).

contra:
Thanks! And yes, we could really use some sourcemaps tests

Issue #137, opened 2015-12-18
The addition of futimes appears to break unbuffered streams


Klowner:
I'm pretty sure this test shouldn't fail, right?
https://github.com/Klowner/vinyl-fs/tree/bug-futimes-hates-buffered-false

Appears to be an issue in `dest/writeContents/writeStream.js`. I assume that `outStream` is actually closed by the time `futimes(outFD, stat, complete)` is called. This may be a graceful-fs bug? I'll look into it further if someone cares to confirm.

phated:
@Klowner this is what I was running into yesterday, couldn't track down the root cause but it is probably closed at that point.

Klowner:
Thanks for confirming @phated.

Looks like it's not an issue with graceful-fs, as normal fs appears to exhibit the same behavior. That is, a WriteStream's associated file descriptor is closed by the time the `'finish'` event is fired. I suppose the easy solution would be to just re-open the output file and call futimes on it, but it would be nice to avoid that.

phated:
Should be closed by the fix in #138

PR #138, opened 2015-12-18
Rework writeStream to prevent issue with futimes

Klowner:
I'm fairly certain this addresses the issue in #137. Includes a unit test.


phated:
Thanks for digging into this and adding a test.  LGTM

Klowner:
:+1: you bet

PR #142, opened 2016-01-12
Sink should respect a readable event on the destination stream

benurb:
Hello :smile: 

by using grunt-contrib-imagemin I encountered a problem that I tracked down to vinyl-fs 2.3.0. If one attaches a listener to the `readable` event of the `vinly-fs.dest` stream, the new sink (introduced in https://github.com/gulpjs/vinyl-fs/commit/df97b1482200e7bc5dd01a49f0bdedeed2b6bd08 ) prevents it from firing. I attached a test to this PR that hopefully explains the problem better.

The problem is that the `pipe()` call in `sink.js` turns the stream into flowing mode. In this mode the stream does no longer emit a `readable` event (see [Node docs](https://nodejs.org/docs/latest/api/stream.html#stream_class_stream_readable)).

This of course is a special case. Nevertheless the destination stream should not be turned into flowing mode if it's obvious that this was not intended. Imho that's the case if a developer adds a `readable` event listener, so I added a check for it.

Feedback is much appreciated! :smile: 

Ben

### DIFF
--- a/lib/sink.js
+++ b/lib/sink.js
@@ -11,6 +11,11 @@ function sink(stream) {
   });
 
   return function() {
+    // Respect readable listeners on the underlying stream
+    if (stream.listeners('readable').length > 0) {
+      return;
+    }
+
     stream.pipe(sinkStream);
   };
 }
--- a/test/dest.js
+++ b/test/dest.js
@@ -1337,6 +1337,29 @@ describe('dest stream', function() {
       .pipe(slowCountFiles);
   });
 
+  it('should respect readable listeners on destination stream', function(done) {
+    var srcPath = path.join(__dirname, './fixtures/test.coffee');
+    var srcStream = vfs.src(srcPath);
+    var destStream = vfs.dest('./out-fixtures/', { cwd: __dirname });
+
+    srcStream
+      .pipe(destStream);
+
+    var readables = 0;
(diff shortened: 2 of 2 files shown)


phated:
This solution was suggested by @chrisdickinson. I'd like to get his feedback on your change.

benurb:
Sure :smile:
The problem currently is that 2.3.0 introduces a breaking change in a minor version update. So I hope we can get this sorted out asap :+1:

phated:
Yep, but a breaking change is intentional, this was a regression. Sorry about that.

benurb:
Not trying to be the semver nazi here, but just expressing my opinion. If you're following the SemVer spec, they are pretty clear on that:

> What do I do if I accidentally release a backwards incompatible change as a minor version?

> As soon as you realize that you've broken the Semantic Versioning spec, fix the problem and release a new minor version that corrects the problem and restores backwards compatibility. Even under this circumstance, it is unacceptable to modify versioned releases. If it's appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.
(Source: http://semver.org)

Imho the related commit should be reverted and a new 2.3.2 should be released. The fix should land in 3.0.0. But of course that's up to you to decide :+1: 

Regards,
Ben

Issue #103, opened 2015-09-29
Possible problem with tests running in a setgid dir


OnkelTem:
These tests fail on my system (Linux/Ubuntu):

1) dest stream should write directories to the right folder
2) dest stream should use different modes for files and directories
3) symlink stream should use different modes for files and directories

After some debugging I figured out that all three tests fail due to the project directory (and its subdirectories) permissions which has `setgid` bit set.

To reproduce execute this command:
```
$ find test -type d -exec chmod g+s {} \;
```
and then run the tests.

=
**Whys**

Let's look at the first failing test `should write directories to the right folder`.
The expected mode of a dir is:
```js
var expectedMode = parseInt('655', 8);
```
and the failing line is:
```js
realMode(fs.lstatSync(expectedPath).mode).should.equal(expectedMode);
```
and my run-time values were:
```js
fs.lstatSync(expectedPath).mode.toString(8) // 42655
realMode(fs.lstatSync(expectedPath).mode).toString(8) // 2655 != 655 (expectedMode)
```
— i.e. it didn't expect an extra bit set by the filesystem.

Ideas? Can't we just make `realMode()` more loosing like this:
```patch
var realMode = function(n) {
-  return n & parseInt('7777', 8);
+  return n & parseInt('777', 8);
};
```

contra:
PR?

PR #144, opened 2016-01-14
Feature metadata

erikkemperman:
Squashed as [requested](https://github.com/gulpjs/vinyl-fs/pull/128#issuecomment-171477651).


phated:
@erikkemperman it seems that `fchmod` can't be used on a file that you don't own.  Can you add the same check as `getTimes` to compare the `uid` and a test to confirm it is working?

erikkemperman:
At a glance, there is no test for the check in `getTimes` yet, either? I will have some time for this tomorrow.

erikkemperman:
Several changes, including the guard against calling `fchmod` when we don't own the file. I have no idea how to unit test this though, we're not allowed to chown a file in order to verify this guard works. Ideas?

The last couple of commits are smaller changes I think are improvements, but for the moment I'll leave them unsquashed for easier review.

erikkemperman:
Note, there are of course still those two tests that fail against the version of `vinyl` with the contents getter/setter changing `stat.xtime`. Because that change is already in `master`, I propose to get that sorted out first -- i.e. in the corresponding [PR](https://github.com/gulpjs/vinyl-fs/pull/143) -- before contemplating this one.

phated:
@erikkemperman we don't rely on the master branch of vinyl so I don't know why they would be failing

phated:
@erikkemperman can you revert some of the changes that were added to handle the vinyl changes I've now removed.  I'd like to get this in but it got a lot more complicated with those workarounds.

erikkemperman:
@phated I've tried to keep the changes related to the now removed `vinyl` changes in a separate PR, [here](https://github.com/gulpjs/vinyl-fs/pull/143). If there's anything in here you'd rather not have I can of course remove those, but I believe these commits are orthogonal to the `vinyl` stuff.

phated:
@erikkemperman ah, I had thought the last 4-5 commits were addressing things from the vinyl changes.

erikkemperman:
@phated No they're just tweaks on @piranna and my efforts that I kept separate for the moment. I think they are slight improvements, especially the commit that makes the `stat` post-dest reflect the file on disk, which ought to keep piping onto `dest` equivalent to having a new `src` on that file.

Note however that the other PR, which is now mostly pointless until the `vinyl` issue is resolved, does contain as requested a more or less unrelated commit to address the intermittently failing tests, [here](https://github.com/erikkemperman/vinyl-fs/commit/b70bed0c7c488c9de6a530e1668aba9a852f2e4e).

I didn't have any time the last week or so, but still think the `vinyl` issue deserves some thought, would love some feedback on that issue.

phated:
@erikkemperman it definitely needs more thought and discussion.  However, can you submit a separate PR with the fix to the failing tests?  I'd like to get that merged.

lib/dest/writeContents/index.js
```diff
@@ -29,49 +41,124 @@ function writeContents(writePath, file, cb) {
...
+  function stat(fd, close) {
+    fs.fstat(fd, function(err, stat) {
+      if (err) {
+        return close(err, finish);
+      }
+
+      // Check if mode needs to be updated
+      var modeDiff = 0;
+      if (typeof file.stat.mode === 'number') {
+        modeDiff = (file.stat.mode ^ stat.mode) & MASK_MODE;
```
phated: what is the `^` operator? I've never encountered it before

phated: nvm, found it at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_XOR

erikkemperman:
Sure thing, will do that when I get on campus in an hour or so.

As to the commits here, I kept them separate for easier review but should probably squash at some point.

lib/dest/writeContents/index.js
erikkemperman: Heh, I guess I betrayed my C origins there. It's a nice operator to get bit-difference.

lib/dest/writeContents/index.js
```diff
@@ -1,11 +1,23 @@
...
+var fs = require('graceful-fs');
+var assign = require('object-assign');
+
 var writeDir = require('./writeDir');
 var writeStream = require('./writeStream');
 var writeBuffer = require('./writeBuffer');
 var writeSymbolicLink = require('./writeSymbolicLink');
 
+// TODO include sticky/setuid/setgid, i.e. 7777?
+var MASK_MODE = parseInt('0777', 8);
```
phated: @erikkemperman  what would happen if this was just `'777'` like we did at https://github.com/gulpjs/vinyl-fs/blob/master/test/dest.js#L38

phated: see https://github.com/gulpjs/vinyl-fs/issues/103 and https://github.com/gulpjs/vinyl-fs/pull/134

erikkemperman: @phated Not sure what you mean... If you meant dropping the zero, so it would leave `parseInt('777', 8)` -- that would do the exact same thing, since it is the same number...

I added the leading zero to emphasize we are now actively ignoring any sticky/setuid/setgid bits that users might have set on `stat.mode`. Notice that this does not set those bits on extant files on disk to 0 -- all this mask does is ignore any bits on `stat.mode` that are not rwx flags. And the TODO comment questioned if we wanted to allow users to set those other bits as well.

Either way, it is separate from the issue you linked, although if we were to allow these bits to be set by the user, and reflected on disk by `dest`, then there should be a test for it which doesn't use the `realMode` function of course (awkward description of what it does, acually).

erikkemperman:
@piranna The fix for the intermittently failing tests would have to be altered again if this PR were merged. So I would propose getting this out the way first and then addressing those?

Let me know if I should remove any commits from this PR, or rebase/squash.

piranna:
> EDIT: Sorry @piranna, auto-complete got me. I meant to address @phated.

Ok, don't worry :-)

erikkemperman:
@phated Done! Thanks for being patient, this whole thing took a good while. Once merged I will adapt and submit the fix for intermittently failing atime/mtime tests.

phated:
@erikkemperman one more thing, we are writing a lot of functions that take the file descriptor *and* the vinyl file object - what do you think about attaching the file descriptor to the vinyl object while we are passing it through the vinyl-fs machinery and then removing it when done?  This would be very similar to how node's fs.WriteStream works

erikkemperman:
@phated I inlined some of those functions, like `times` and `mode`, so they would have some of their operands in scope and not have to be passed everything. But that won't work across module boundaries obviously.

So you're suggesting to have the various `writeXXX` modules do `file.fd = ...` before calling the `written` callback, and then to have each of them do `delete file.fd` in their `close` callback?

I guess I don't see why not, but then I also don't really mind a separate argument in what is after all an internal contract.

erikkemperman:
@phated Would it be a completely ludicrous idea to de-couple the opening/closing of filedescriptors, i.e. have a separate utility which handles this stuff? That might do lazy closing or even predictive opening? I imagine it is pretty common to have intermediate (but persistent) files, and such cases might benefit from subsequent pipes discovering that its file still has a handle.

phated:
Finished by #151 - Thanks for all your work on this @erikkemperman!

PR #145, opened 2016-01-14
Allow things to listen for data or readable events and avoid sinking in those cases

phated:
Closes #142 

@chrisdickinson singular diff, what do you think?


phated:
I believe this to be a complete implementation of the solution.  It has been tested thoroughly in our test suite and I've linked it into the grunt-contrib-imagemin project to run their test suite and everything is passing.  Going to merge and release as 2.3.1

PR #151, opened 2016-01-28
Feature metadata refactor

phated:
Clean diff against master.  /cc @erikkemperman @contra @piranna 

Ref #144


contra:
@phated Got a brief summary of the changes inside this? I haven't been keeping up with the symlink/chown issues

phated:
@contra Goals of these PRs:
* Make sure `futimes` is only called if there is a difference between on-disk and in-memory and only if we have the correct permissions to actually do it (aka own the file)
* Make sure `fchmod` is only called if there is a difference in permissions on-disk and in-memory and only if we have the correct permissions to actually do it (aka own the file)
* Use file descriptors everywhere to reduce the overhead of these operations 
* Update test for the above
* __Still TODO:__ add lots more tests for the new cases

A lot of stuff is boilerplate that would have happened in node core (like my `writeFile` helper module) because we need finer grained control over the file descriptor after the content is written.

erikkemperman:
> I haven't been keeping up with the symlink/chown issues

Actually `fchown` is not in there. Yet? It could write `stat.gid/uid` to disk as well, if we are owner/root, respectively.

phated:
@erikkemperman currently trying to fix the features we currently support, we can talk about feature additions after.

erikkemperman:
Fair enough :-) Just it was cute that `chown` was mentioned, it does belong with chmod and touch, that way.

phated:
Working on some tests today

phated:
@erikkemperman should we use `process.umask()` anywhere in this new code?

phated:
Still more to do but made great progress today.  Now sleep.

erikkemperman:
@phated 
`process.umask()` -- Maybe, if the `mode` on the vinyl is not defined. Currently in [writeFile](https://github.com/gulpjs/vinyl-fs/blob/blaine-feature-metadata-refactor/lib/dest/fileOperations/writeFile.js#L27) we default to '666' (very metal :-)) but we could do `& ~process.umask()` like [mkdirp](https://github.com/substack/node-mkdirp/blob/master/index.js#L20) does. But I would expect the `fs` functions to do this for us, anyway?

PS I would always move constants created with a function call (like `parseInt(..,8)`)  out of the critical path -- unless we can be sure v8 will optimize these?

phated:
@erikkemperman @piranna @contra I've finished the "fileOperations" tests with 100% coverage.  Could you review and let me know if anything is missing?  I plan to work on "dest" tests next to get coverage on the new stuff and handle some edge cases we aren't testing.  Thanks!

phated:
Coverage is through the roof, finished the fileOperations unit tests (as noted above), added some more `dest` tests (including some for unowned files).  @erikkemperman @piranna @contra Please review everything when you get a chance because I think we are closing in on a complete PR.

Once I get thorough reviews, I'll squash my stuff together.

Anyway, tests are exhausting so I'm headed off for the night.

lib/dest/index.js
```diff
@@ -32,7 +32,7 @@ function dest(outFolder, opt) {
 
   var sourcemapOpt = opt.sourcemaps;
   if (typeof sourcemapOpt === 'boolean') {
-    sourcemapOpt = { sourcemaps: sourcemapOpt };
+    sourcemapOpt = {};
```
phated: this wasn't a valid option for gulp-sourcemaps so I snuck it in

erikkemperman:
@phated Unfortunately I am kinda swamped this week -- will have some time to review more thoroughly this weekend if still needed.

Couple of things I notice at a glance:

I was really curious how you would go about unit testing the not-owned file checks. Because I am pretty sure that files you get from git are always owned by whoever ran the git command that put them there. How could it not?

And sure enough, having just fetched in your branch, the `not-owned` file is owned by me. But that means that those tests might not be doing what they should?

Another thing I notice is that there are a couple `should(<spy>.called).be.ok` and `...be.not.ok` tests in there, which I discovered a couple days back doesn't actually appear to ever make a test fail. See the related [PR](https://github.com/gulpjs/vinyl-fs/pull/155). So I would guess something like that would have to be changed within this PR as well.

About whether or not to throw errors on failed `fchmod` or `futimes` calls... We've talked about that early in this whole affair. @piranna 's implementation just skips silently any call that it determines wouldn't be allowed (ownership). My refactor, and I guess yours on top of that, does the same. But if folks make a point of setting a custom `mode`, or `atime/mtime` on the vinyl object, they might want to know about failure to write those attributes.

Would it therefore  make sense to have something like a `pedantic` option which, if set, would either
- throw an error if we determine `fchmod` or `futimes` will fail, or
- skip the test for ownership and just attempt the call/s, propagating up any errors off it/them?

phated:
@erikkemperman thanks for the quick points and I'll wait on any merges until I get a thorough review. I should have time to work on this on the weekend also.  I'll address your comments below:

Thanks for checking the ownership stuff.  I didn't know that's how git worked because it keeps file modes and it would have been annoying for me to check that.  I'll dig into it some more.

I don't believe I have added any lines with `should(spies.called)` but there are some in the diff so it must be somewhere in our rebases.  I will make the changes like your other PR.  On a side note, the new tests I've added are using `expect()` instead of `should` and I want to do a pass on all the tests and swap it completely (in a separate PR after this).

I don't think it's appropriate to swallow the errors, especially now that we have a lot of checks on the diff and ownership; However, if the vinyl object is a directory and we successfully write it with mode `000`, then we aren't going to be writing anything inside it (and if we did, that would error) so I don't see the point in erroring on that open.  I guess if they wrote the directory as `000` but then wanted to update the mtime/atime, that would silently succeed but not update the times.

erikkemperman:
@phated 
Well, since the git remote doesn't (shouldn't!) know about my local uid/gid mapping it kinda makes sense. Modes are universal that way.

> I don't think it's appropriate to swallow the errors

So instead of having a `pedantic` option and defaulting to silently skipping `futimes` and `fchmod`, you would favour throwing (most) errors instead and possibly a `lenient` option to get the current behaviour? 

I actually think that makes sense for unlikely edge-cases like directory mode '000' too -- it won't error until user tries to violate his own restriction, which is not vfs' problem?

erikkemperman:
Hm, you were probably talking about the order in which `fchmod` and `futimes` are called. For a directory vinyl object with mode `000` the `fchmod` might work but then the `futimes` would fail. We would want to do them in reverse order.

But conversely, if the directory already exists and the mode on the vinyl is less restrictive than the one on disk, then `futimes` might fail if we run it before the `fchmod`, and the so order we have now is what we want.

Should we be determining whether the requested `fchmod` makes for a more or a less restrictive restrictive subject for `futimes`, and adjust the call order accordingly?

phated:
@erikkemperman the mode would be the same (barring any crazy getter usage) because the mkdirp call is passed the `stat` object from the vinyl object.  The only thing that could be different would be the atime/mtime, which wouldn't be called.  The only way to make this work would be to check the diff before calling open but if we don't open it, we can't use `fstat`.  I wonder if `fs.stat` works if permissions are `000`

erikkemperman:
No time to check just now, but does mkdirp also update the mode from `file.stat.mode` if the directory pre-exists?

phated:
@erikkemperman I don't think it updates it if it already exists.  I went down a rabbit hole and looked through open(2) manpages;  it seems there is a constant (O_PATH) that lets you open a descriptor for things like `fclose` and `fstat` but fails on reading/writing/etc.  However, that constant doesn't exist in node's `constants` module so I'm not sure what to do.

erikkemperman:
And I think node's documentation of such things is... Slightly Spartan for my taste.

But I think there might be different optimal orders in which to execute `futimes` and `fchmod`, depending on both the vinyl object and the mode of the physical file if it already exists.

phated:
I think everything with the updateMetadata is fine.  I am currently thinking we need to remove the `mkdirp` call from the `writeDir.js` file and use `fs.open` on the directory we are creating before creating it so the open can't fail based on permissions.  If a directory already exists there and we don't have permissions to create one, then we let the error happen.

phated:
@erikkemperman doing some experiments with the `not-owned` files and it looks like the only way to have tests for this is to use travis-ci's non-container infrastructure and then use sudo to chown the files to root.  I can skip the tests on local development and only run on Travis with a nice log message in local dev.  This is the best I can do, it seems.

phated:
@erikkemperman hoping you have some time this weekend for a thorough review. Would like to get this merged soon.

erikkemperman:
@phated Did not find the time I had hoped for today, and looks like tomorrow's unexpectedly filled up as well. Sorry about that. Did spend a little while on it earlier today, though, and nothing really jumped out at me thus far.

Some minor things:

- I understand your point about separating stuff to facilitate unit testing, but it all still feels a bit fragmented to me... And I've spent some time with this stuff by now -- it might be slightly bewildering to fresh eyes. Perhaps there is some middle ground option, say with a multiple-export module.
- the `MASK_MODE` constant should perhaps be widened to include setuid/setgid/sticky bits, as the TODO comment says. I kept this value from piranna's effort, which in turn was probably inherited from before; but actually I can't think of a good reason to disallow these bits if users bother to set them on the vinyl.
- I'm not really sure to what extent we can assume v8 to apply certain optimizations, but to be on the safe side maybe it is better to not have conditions like
```js
if (typeof process.geteuid === 'function') {
   uid = process.geteuid();
 } else {
   uid = process.getuid();
 }
```
in the main path (I put it there, I know). It will return the same value each and every time, so an `if` checked just once during initialization would seem to reflect that better. Something similar applies to things like `parseInt(...,8)`, which it seems to me should be "constants" (as was done with `MODE_MASK`).

More to the point though, it looks to me like this PR does the job just fine. Perhaps @piranna could also take a look? Much of this began with his hard work!

For posterity I might throw in some thoughts I had along the way, just in case they're not completely silly:
- apply uid/gid from the vinyl to the file on disk (`fchown`)?
- decouple opening and closing of filedescriptors from `src` and `dest`, i.e. in a separate utility that might close descriptors lazily (having it ready in case a subsequent pipeline wants it)?
- and of course there is still the open issue of whether/how `vinyl` should manage the stat contents during a virtual file's lifetime.

erikkemperman:
Oh, and the question of throwing vs suppressing failure to reflect the vinyl's mode and/or times... Maybe I misunderstood, but I thought the idea was now to not swallow most errors... However, the current diff still returns quietly when we're not allowed to call `fchmod` / `futimes`?

phated:
@erikkemperman I'm going to get this wrapped up today.

* I merged the fileOperations into a single file.  I dislike this style because it makes the file huge but I'm willing to make the compromise if the individual modules were hard to follow.
* I don't think we should deal with widening the MASK_MODE in this PR.  I'll create an issue for it after the merge.
* V8 should optimize that example fine but we actually need to call it each time because a user can call `process.setuid()` at any point in their program to change it.  I'll hoist any defaults that use `parseInt` or similar.

All your points under "for posterity" will get created as issues once this is merged.  Thanks for keeping track of them so far.

You are correct that we don't want to swallow the majority of errors, but I think that is different than checking the owner and not attempting.  We want to intelligently determine when it makes sense to even attempt the `fchmod` or `futimes` in the first place;  If we aren't allowed to update them, we don't try and don't error (fixes common use cases as per https://github.com/gulpjs/gulp/issues/1502 and https://github.com/gulpjs/gulp/issues/1376, etc) but if we are allowed and it errors, then we propagate the error.  I think this is the best/most suitable solution to all the problems.

phated:
Looks like the tests don't run on Windows :sob: Glad I decided to turn that on for this PR

erikkemperman:
@phated Actually I meant the test if we have `process.geteuid` or must revert to `.getuid`, not the user id itself!

About the error handling, sounds fine. I maintain that it might be useful to be able to make this error (if I go out of my way to set an exotic mode I'd be sad if `dest` silently skips applying it) but not by default and so not necessarily in this PR.

Good to see this nearing the finish line, thanks a lot!

erikkemperman:
Windows.. The torture never stops.

phated:
Well this is depressing: https://github.com/nodejs/node-v0.x-archive/issues/4812

erikkemperman:
Oh dear, that _is_ depressing. The issue you linked is 3 years old and looks like they won't (can't) fix.

It mentions Windows 7, but I imagine the same problem happens for later versions as well.

I'm curious if this happens with cygwin as well, but I suppose it's academic.

phated:
@erikkemperman my series of commits document my nightmares today.  I'm signing off for the night but I got it down to 5 failing tests.  Those 5 look like it will be really hard to track down the problems.

erikkemperman:
@phated Looks like you had a real fight with it... I have no idea of their numbers in gulp land, but sure the windows crowd is going to appreciate it!

phated:
@erikkemperman found some really annoying problems with Windows so I'm actually disabling the `updateMetadata` function on Windows by always returning false if `getuid` or `geteuid` are both undefined.  The majority of vinyl-fs will continue to work but they won't have access to this feature.  I have plans on a possible way to support this in the future.

erikkemperman:
@phated That sounds perfectly reasonable.. But then I don't use Windows so wouldn't be affected either way.

phated:
:+1: me neither but I think having the feature would cause more problems that it solves... It turns out you can't call futimes on a directory in Windows but can on files...

phated:
Squashed my windows test fix commits but it seems like node 0.12 and 5 are failing for different reasons.  0.10 and 4 pass...

phated:
Wow, a green check next to appveyor...it's the beginning of the end!

phated:
Rebased and tests are running.  Going to review the full diff and write some docs on these changes (and note it doesn't work on Windows) and then ship this.

phated:
:shipit:

phated:
Thanks for everyone that helped out on this: @piranna @erikkemperman @contra!  This has been shipped as 2.3.2

contra:
@phated Thanks for taking this across the finish line 🌴

Issue #153, opened 2016-02-06
Quirk in passing `read` option from `src` to `through2`?


erikkemperman:
It is unlikely to cause any real trouble, but thought I'd report this anyway:

Wondering if I could provide function values for `src` options like `read` and `buffer`, I just tried my luck. It failed, but in a mildly interesting way...

Turns out that, since the `src` options are passed on to `through2`, I was actually telling the readable streams to invoke my `read` option!

Now I don't know if it actually makes sense to have function values for such options, but I figured that because it means quite different things for `src` and for `through2`, it might be better to not pass this `read` option from one into the other?

And looking at the [commit](https://github.com/gulpjs/vinyl-fs/commit/d553e7474ebb65b89e496dfc3916292b59eec4c2) that introduced this, I am also curious why it passes `opt` instead of the sanitized `options`?

phated:
Just an oversight. Can you create a PR with a failing test and possibly a fix?

erikkemperman:
Probably tomorrow. And would it be a ludicrous suggestion to allow function values?

phated:
I'm still working on that, I'd like to get it in after our fix release that I'm writing tests for.

erikkemperman:
Hm, maybe I had the wrong idea. Thought of this function value of `src.options.read` being invoked for each vinyl object, but people probably expect such an option function to apply to all of them, just like when the option is boolean?

erikkemperman:
PR merged, thanks. Closing.

phated:
Published as 2.4.1 - thanks again @erikkemperman

PR #154, opened 2016-02-06
Don't pass `read` option from `src` to `through2`

erikkemperman:
As promised in [#153](https://github.com/gulpjs/vinyl-fs/issues/153)

### DIFF
--- a/lib/src/index.js
+++ b/lib/src/index.js
@@ -43,6 +43,10 @@ function src(glob, opt) {
     followSymlinks: true,
   }, opt);
 
+  // Don't pass `read` option on to through2
+  var read = options.read !== false;
+  options.read = undefined;
+
   var inputPass;
 
   if (!isValidGlob(glob)) {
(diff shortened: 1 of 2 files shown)


phated:
@erikkemperman can you add a test for this?  I'm not sure the behavior passing it caused

erikkemperman:
I'm not sure what it would test...? What I noticed is if I pass a function to the `read` option of `src` -- expecting its boolean return value to be used to determine whether or not to read the file contents -- it was instead being treated as the `read` implementation of stream, because it is passed on to `through2`.

So I suppose I could write a test that does that and check that the function is not called (that my PR makes sure it's not passed to through2). But you mentioned that function values are going to be supported soon, so then this test would be in the way.

Anyway, the diff is pretty self-evidently doing what it advertises, I thought?

phated:
Yeah, I just would like some sort of test so I don't add a regression during a refactor. I actually tried refactoring my other PR today but all the tests caught my mistakes and I realized I couldn't do what I was trying. I'd be happy with the test suggestion you made, it shouldn't be in the way.

erikkemperman:
All right, will do. I'll add a comment though because I do expect this test might confuse whoever ends up implementing function-value support for options like `read` and `buffer`.

erikkemperman:
@phated Rebased and added the requested test, with comment on how to tweak it if/when function-values for options like `read` are supported. You can make it fail by removing line [48](https://github.com/erikkemperman/vinyl-fs/commit/3b67921fe5ddd925fc12acef2bd03777b0081dc3#diff-2599ed3c945bd61f0161c2419a569f0bR48) from src/index.js.

phated:
Thanks @erikkemperman!  Published as 2.4.1 (those travis tests took forever)

PR #155, opened 2016-02-06
Fix spies

erikkemperman:
Another little thing I just noticed... I accidentally had

```js
should(chmodSpy.called).be.ok;
should(chmodSpy.called).be.not.ok;
```

both in a test -- and it passed. It appears to have to be an actual boolean compare.

Also there were some redundant `reset()` calls, these are handled in setup/teardown.

### DIFF
--- a/test/dest.js
+++ b/test/dest.js
@@ -446,7 +446,6 @@ describe('dest stream', function() {
       done();
     };
 
-    chmodSpy.reset();
     var stream = vfs.dest('./out-fixtures/', { cwd: __dirname });
 
     var bufferStream = through.obj(dataWrap(buffered.push.bind(buffered)), onEnd);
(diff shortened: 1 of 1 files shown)


phated:
@erikkemperman not sure if these are still needed or if I fixed them in my changes.  Mind rebasing and checking?  Thanks.

erikkemperman:
Yes, sure, I'll do that when I get on campus. You mentioned you were using a different assert lib in the metadata stuff, so I guess I'll use the same here?

phated:
If you want to. We have a mixture now and I plan to update the old tests.

erikkemperman:
@phated After rebase my commit was empty: all the tests containing wrong spy checks were removed from `test/dest.js`, and the new tests in `destModes.js` and `destTimes.js` don't have any problematic checks because they use `expect` instead of `should`.

Ergo, closing this PR.

PR #163, opened 2016-04-20
changed source file names to kebab case (fix #152)

### LINKED ISSUE #152: kebabCase the filenames
Case-insensitive file systems are a pain.  I think we should kebabCase the filenames like `write-buffer.js`.  I would consider this to be a breaking change.

hgwood:
Pretty straight forward :). Easy to make a mistake though, but I guess the coverage is good enough to catch `require` errors.


phated:
Rebased locally and pushed as 5cf7de1df6fc47886aaa72c1737490069e50ab3b

Thanks!

PR #167, opened 2016-04-26
Support setting setuid/setgid/sticky in updateMetadata

tmcgee123:
Adapt tests to support for setuid/setgid/sticky bits
Relates to #156


phated:
@tmcgee123 it says that no files have changed.  I think this is already in master.  Note that I merge with rebase now so you commit SHAs might differ (looks like you merged the master rebase)

PR #168, opened 2016-05-02
Support functions for all src options (fix #139)

### LINKED ISSUE #139: Support functions for all options
Ref https://github.com/gulpjs/gulp/issues/1450

hgwood:
It is not entirely clear from the issue if it is only about `src` or if `dest` is also in scope. Anyway, here it is for `src`. I can replicate the behavior for `dest` very simply.

A few notes:
- I think mixing vinyl-fs options and through2 options in the option object can be error-prone (https://github.com/gulpjs/vinyl-fs/issues/153) and can complexify the implementation (see how I need to avoid calling through2 options that are functions?). May I suggest moving through2 options to a through2 property in vinyl-fs options? I believe that would be simpler and clearer, both for the caller and the callee. It would be a breaking change.
- I did not use `value-or-function` because I did not see the benefit of checking the types of the options. Also I saw defaulting to `null` as a bit dangerous and surprising for users of vinyl-fs.
- What about jumping to Node 4 LTS compatibility for vinyl-fs 3? That would enable a bunch of ES6 features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,21 +61,8 @@ fs.src(['*.js', '!b*.js'])
 
 #### Options
 
-##### `options.cwd`
-
-The working directory the folder is relative to.
-
-Type: `String`
-
-Default: `process.cwd()`
-
-##### `options.base`
-
-The folder relative to the cwd. This is used to determine the file names when saving in `.dest()`.
-
-Type: `String`
-
-Default: The part of the path before the glob (if any) begins. For example, `path/to/**/*.js` would resolve to `path/to`. If there is no glob (i.e. a file path with no pattern), then the dirname of the path is used. For example, `path/to/some/file.js` would resolve to `path/to/some`.
+- Values passed to the options must be of the right type, otherwise they will be ignored.
+- All options can be passed a function instead of a value. The function must return a value of the right type, otherwise it will be ignored.
 
 ##### `options.buffer`
 
@@ -137,6 +124,7 @@ Default: `true`
 
 Any glob-related options are documented in [glob-stream] and [node-glob].
 Any through2-related options are documented in [through2].
+Those options are forwarded verbatim.
 
 ### `dest(folder[, options])`
 
@@ -156,6 +144,9 @@ __Note: The file will be modified after being written to this stream.__
   - `contents` will have it's position reset to the beginning if it is a stream.
 
 #### Options
+
+- Values passed to the options must be of the right type, otherwise they will be ignored.
+- All options can be passed a function instead of a value. The function will be called with the [vinyl] `File` object as its only argument and must return a value of the right type for the option.
 
 ##### `options.cwd`
 
@@ -232,6 +223,7 @@ Default: `undefined` (do not write sourcemaps)
 ##### other
 
 Any through2-related options are documented in [through2].
+Those options are forwarded verbatim.
 
 ### `symlink(folder[, options])`
 
@@ -242,6 +234,9 @@ __Note: The file will be modified after being written to this stream.__
   - `cwd`, `base`, and `path` will be overwritten to match the folder.
 
 #### Options
+
+- Values passed to the options must be of the right type, otherwise they will be ignored.
+- All options can be passed a function instead of a value. The function will be called with the [vinyl] `File` object as its only argument and must return a value of the right type for the option.
 
 ##### `options.cwd`
 
@@ -270,6 +265,7 @@ Default: The process mode.
 ##### other
 
 Any through2-related options are documented in [through2].
+Those options are forwarded verbatim.
 
 [glob-stream]: https://github.com/gulpjs/glob-stream
 [gulp-sourcemaps]: https://github.com/floridoo/gulp-sourcemaps
--- a/lib/dest/index.js
+++ b/lib/dest/index.js
@@ -3,6 +3,8 @@
 var through2 = require('through2');
 var sourcemaps = require('gulp-sourcemaps');
 var duplexify = require('duplexify');
+var valueOrFunction = require('value-or-function');
+
 var sink = require('../sink');
 var prepareWrite = require('../prepare-write');
 var writeContents = require('./write-contents');
@@ -11,6 +13,9 @@ function dest(outFolder, opt) {
   if (!opt) {
     opt = {};
   }
+
+  var sourcemapsOpt = valueOrFunction(
+      ['boolean', 'string', 'object'], opt.sourcemaps);
 
   function saveFile(file, enc, cb) {
     prepareWrite(outFolder, file, opt, function(err, writePath) {
@@ -22,7 +27,7 @@ function dest(outFolder, opt) {
   }
 
   var saveStream = through2.obj(opt, saveFile);
-  if (!opt.sourcemaps) {
+  if (!sourcemapsOpt) {
     // Sink the save stream to start flowing
     // Do this on nextTick, it will flow at slowest speed of piped streams
     process.nextTick(sink(saveStream));
@@ -30,17 +35,15 @@ function dest(outFolder, opt) {
     return saveStream;
   }
 
-  var sourcemapOpt = opt.sourcemaps;
-  if (typeof sourcemapOpt === 'boolean') {
-    sourcemapOpt = {};
-  }
-  if (typeof sourcemapOpt === 'string') {
-    sourcemapOpt = {
-      path: sourcemapOpt,
+  if (typeof sourcemapsOpt === 'boolean') {
+    sourcemapsOpt = {};
+  } else if (typeof sourcemapsOpt === 'string') {
+    sourcemapsOpt = {
+      path: sourcemapsOpt,
     };
   }
 
-  var mapStream = sourcemaps.write(sourcemapOpt.path, sourcemapOpt);
+  var mapStream = sourcemaps.write(sourcemapsOpt.path, sourcemapsOpt);
   var outputStream = duplexify.obj(mapStream, saveStream);
   mapStream.pipe(saveStream);
 
--- a/lib/prepare-write.js
+++ b/lib/prepare-write.js
@@ -5,39 +5,34 @@ var path = require('path');
 var mkdirp = require('mkdirp');
 var fs = require('graceful-fs');
 
-function booleanOrFunc(v, file) {
-  if (typeof v !== 'boolean' && typeof v !== 'function') {
-    return null;
-  }
+var valueOrFunction = require('value-or-function');
+var defaultValue = require('./default-value');
 
-  return typeof v === 'boolean' ? v : v(file);
-}
-
-function stringOrFunc(v, file) {
-  if (typeof v !== 'string' && typeof v !== 'function') {
-    return null;
-  }
-
-  return typeof v === 'string' ? v : v(file);
-}
+var boolean = valueOrFunction.boolean;
+var number = valueOrFunction.number;
+var string = valueOrFunction.string;
 
 function prepareWrite(outFolder, file, opt, cb) {
-  var options = assign({
-    cwd: process.cwd(),
-    mode: (file.stat ? file.stat.mode : null),
-    dirMode: null,
-    overwrite: true,
-  }, opt);
-  var overwrite = booleanOrFunc(options.overwrite, file);
-  options.flag = (overwrite ? 'w' : 'wx');
+  if (!opt) {
+    opt = {};
+  }
+
+  var defaultMode = file.stat ? file.stat.mode : null;
+  var options = assign({}, opt, {
+    cwd: defaultValue(process.cwd(), string(opt.cwd, file)),
+    base: string(opt.base, file),
+    mode: defaultValue(defaultMode, number(opt.mode, file)),
+    dirMode: number(opt.dirMode, file),
+    overwrite: defaultValue(true, boolean(opt.overwrite, file)),
+  });
+  options.flag = (options.overwrite ? 'w' : 'wx');
 
   var cwd = path.resolve(options.cwd);
-  var outFolderPath = stringOrFunc(outFolder, file);
+  var outFolderPath = string(outFolder, file);
   if (!outFolderPath) {
     throw new Error('Invalid output folder');
   }
-  var basePath = options.base ?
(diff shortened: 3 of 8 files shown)


phated:
@hgwood thanks for submitting this.  I believe we wanted the logic to be applied to `.dest` and `.symlink` options.

* Wouldn't we want to resolve the through2 options that are a function?  Since the bug was caused by `read` being a function, it should always be resolved to a value. 
* `value-or-function` should be used because we only want options to accept a certain type.  I'm also noticing that you missed the fact that each function-option should be called with the vinyl file object (one of the primary reasons `value-or-function` exists).
* We are sticking with backwards compat to node 0.10 because it is heavily used (myself included).

hgwood:
> I believe we wanted the logic to be applied to .dest and .symlink options.

Will do.

> you missed the fact that each function-option should be called with the vinyl file object

I meant to forward the glob to function-options but forgot. Is the glob what you mean by "vinyl file object"? Otherwise could you specify what you mean? I don't see anything in `src()` that is available before options are used.

> Wouldn't we want to resolve the through2 options that are a function? Since the bug was caused by read being a function, it should always be resolved to a value.

Just to be sure we share the same understanding:
- `read` is a vinyl-fs option expecting a boolean.
- `read` is also a through2 option ([`stream.Readable`](https://nodejs.org/api/stream.html#stream_new_stream_readable_options) option actually) expecting a function.
- vinyl-fs resolves this ambiguity by always considering the `read` option is meant for itself. It *never* forward the `read` option to through2.
- There are other through2 options that expect a function, like [`write` and `writev`](https://nodejs.org/api/stream.html#stream_new_stream_writable_options).

If vinyl-fs supported passing function-options for through2 options then how could it distinguish between someone trying to pass a function to through2 and someone trying to pass a function-option that returns a function for through2? I think this could become very confusing for users. So I was careful to stay clear of this and my implementation only calls function-options for vinyl-fs own options.

> we only want options to accept a certain type

This seems to clash with how JS usually handle things. When I read a documentation for a JS package and it says it expects a boolean, I expect that I can pass anything and it will be coerced to a boolean. I expect `{read: "read"}` to mean `{read: true}` while `value-or-function` would interpret that as `{read: false}`. Might be just me though.

Anyway that's not, as you know, how vinyl-fs currently behaves. All options are type-checked already. Boolean options are checked using `option === true`, and the only non-boolean option (`since`) is type-checked using `typeof` and `instanceof` in `filter-since.js`. As a result, adding `value-or-function` on top would amount to no gain. Unless all current type-checks are removed of course.

Quick pros and cons of `value-or-function` over current:
- Pro: *some* (not all) type-checking throughout the program may be removed
- Cons: still have to type-check for "subtypes" of `object` (Date, RegExp, ...) (e.g. `since`)
- Cons: have to maintain a option-to-type map
- Cons: introduces complexity when checking an option for multiple possible types  (e.g. `since`)
- Cons: introduces questions like "should `filterSince` do its own type-checking or should it rely on its parameter being an option that has been type-checked"?

That said, this is pretty much up to personal preference. You are the maintainer of this project and you decide. :)

contra:
@phated Any reason why we shouldn't apply this across the board? (src, dest, and symlink)

hgwood:
I've applied the behavior to `dest` and `symlink`. Function-options for `src` are passed `glob` and function-options for `dest` and `symlink` are passed `file`, except for `dest`'s `sourcemaps` option, which is resolved prior to getting access to `file`.

contra:
@hgwood Assuming this looks good to @phated would need docs updates before merge

Thanks again for tackling this one! Really appreciate it 🙏

hgwood:
I have noticed `booleanOrFunc` and `stringOrFunc` in `prepare-write.js`. Their behavior matches `value-or-function` (returns `null` if the type's wrong). I don't want to break this behavior so now I have a good reason to use `value-or-function`. I'll refactor.

phated:
@hgwood correct, `value-or-function` was written based on the behavior already in vinyl-fs as a way to normalize all the options.  Please use that in this implementation.  If you need other functionality in that module, feel free to submit a PR and I'll review.

hgwood:
The new proposal leverages `value-or-function`. Thank you for you patience as I get more familiar with the code base. :)

I guess the only significant remaining question is: do we agree on the fact that we should only process options that are for vinyl-fs itself? See the second point in [my earlier comment](https://github.com/gulpjs/vinyl-fs/pull/168#issuecomment-216663834).

phated:
@hgwood I'm not really happy with the colliding options, so maybe our `read` option needs to be named something else.  I don't really like the idea of having them as separate keys (e.g. `options.through2` or similar).  I'd like to hear @contra and @erikkemperman's thoughts.

hgwood:
@phated Sorry I think I might have confused you. The colliding option is actually not the problem.

You said:

> Wouldn't we want to resolve the through2 options that are a function?

The thing is, some through2 options *are* functions (namely `read`, `write`, `writev`). So how can vinyl-fs tell the difference between a function that should be passed verbatim to through2, or a function that is a function-option to be passed to `value-or-function`? I don't see how it can. That's why my proposal *do not* process options that are not known to vinyl-fs itself.

phated:
@hgwood The `value-or-function` module already handles function options using the `.function()` method (which would always result in true for a function so it wouldn't resolve the function), but that isn't the actual problem.  The actual problem is that we have a `read` option that is a boolean but it collides with through2's `read` option that is a function.

hgwood:
Oh, right. The `.function()` thing was obvious, sorry.

erikkemperman:
Renaming the `read` option would work, but I guess I would also prefer having a separate namespace for the through2 options. For one thing, I think renaming will be inconvenient for lots of people. Also, what if future versions of through2 and/or vinyl-fs introduced other colliding option keys?

phated:
Separate namespaces get really messy, as we pass options down through glob-stream into node-glob and so forth. Would they be objects nested in objects nested in objects? How far does it go? It seems really clunky

erikkemperman:
True, it's not pretty. But eliminates collisions, current or future. Renaming the vinyl-fs option (to what?) is also not exactly elegant, imho.

I don't have a strong preference either way, but since you asked :-)

lib/prepare-write.js
```diff
@@ -1,43 +1,27 @@
...
-  var options = assign({
-    cwd: process.cwd(),
-    mode: (file.stat ? file.stat.mode : null),
-    dirMode: null,
-    overwrite: true,
-  }, opt);
-  var overwrite = booleanOrFunc(options.overwrite, file);
-  options.flag = (overwrite ? 'w' : 'wx');
+  var options = normalizeOptions(opt, {
+    cwd: { type: 'string', defaultValue: process.cwd() },
```
phated: This is awkward.  Would there be a better way to implement it?  Maybe something like `defaultValue(valueOrFunction.string(opt.cwd, file), process.cwd())`?

hgwood: Could you explain why you find it awkward?

As for the alternative solution, do you mean I should duplicate this line for every option? Like:

```
var options = {
  cwd: defaultValue(valueOrFunction.string(opt.cwd, file), process.cwd()),
  mode: defaultValue(valueOrFunction.number(opt.mode, file), (file.stat ? file.stat.mode : null)),
  dirMode: ...
};
```

phated: I feel it's awkward because you invented a DSL for the option + type + default.  Your alternative is kind of what I was thinking.  It might even look cleaner if we "destructured" the methods from `value-or-function`, e.g. `var str = require('value-or-function').string`

hgwood: I understand what you're saying. I'll have another go.

hgwood:
@phated I've refactored the code. Tell me if you think it's better.

lib/src/index.js
```diff
@@ -11,16 +11,22 @@ var isValidGlob = require('is-valid-glob');
...
-    sourcemaps: false,
-    passthrough: false,
-    followSymlinks: true,
-  }, opt);
+  if (!opt) {
+    opt = {};
+  }
+
+  var options = assign({}, opt, {
+    buffer: normalizeOption('boolean', true, opt.buffer, glob),
```
phated: not sure how I'm feeling about the glob being passed in now that I see it.  I don't think it is useful to determine the options.  I'm going to think a bit more on a way to defer option resolution until we have a Vinyl File object.

lib/normalize-option.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = function normalizeOption(types, defaultValue, value) {
+  if (value === undefined) {
+    return defaultValue;
+  }
+  if (typeof types === 'string') {
+    types = [types];
+  }
+  var appliedArguments = [].slice.call(arguments, 3);
+  for (var index in types) {
```
phated: this loop seems very heavy handed to be used in every scenario when it is only accounting for 2 cases (`sourcemaps` and `since`).  Might need to think on this a bit more.

lib/normalize-option.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var valueOrFunction = require('value-or-function');
+
+module.exports = function normalizeOption(types, defaultValue, value) {
```
phated: I like function composition more than a lot of arguments.  I prefer 1 or 2 arguments (3 max) but the applied arguments make the calls longer.

lib/dest/index.js
```diff
@@ -1,17 +1,24 @@
...
 var prepareWrite = require('../prepare-write');
+var normalizeOption = require('../normalize-option');
 var writeContents = require('./write-contents');
 
 function dest(outFolder, opt) {
   if (!opt) {
     opt = {};
   }
 
+  assign(opt, {
```
phated: this is mutating the `opt` object, which is a something we don't want to do.  You can probably just assign to a `var sourcemapsOpt = ` if it isn't used anywhere else.

hgwood: I'll use the same technique as in `src`.

lib/normalize-option.js
hgwood: Right. I can shortcut it in the second `if`.

lib/src/index.js
hgwood: Removed `glob`. Deferring to get the file can be a PR on its own I think.

lib/normalize-option.js
hgwood: The best would be something like `typed(valueOrFunction(defaultValue(option, defaultValue), 'string', 'object')`. The fact that the `value-or-function` does both "value or function" and "typed" without supporting multiple types is limiting.

phated: yeah, the only problem with having `value-or-function` not aware of types is that you could never have a `function` typed option.  Maybe it just needs to be expanded to support an array of types (and handles the common case of a singular type in the fast path).

phated: Or maybe `value-or-function` should support a function along with strings as the first argument; if we did that, we could properly support `Date or timestamp` as the type for `since`.  Thoughts?

hgwood: Sounds quite elegant.

hgwood: PR for `value-or-function`: https://github.com/gulpjs/value-or-function/pull/1.

README.md
```diff
@@ -61,21 +61,8 @@ fs.src(['*.js', '!b*.js'])
...
-
-##### `options.base`
-
-The folder relative to the cwd. This is used to determine the file names when saving in `.dest()`.
-
-Type: `String`
-
-Default: The part of the path before the glob (if any) begins. For example, `path/to/**/*.js` would resolve to `path/to`. If there is no glob (i.e. a file path with no pattern), then the dirname of the path is used. For example, `path/to/some/file.js` would resolve to `path/to/some`.
+- Values passed to the options must be of the right type, otherwise they will be ignored.
+- All options can be passed a function instead of a value. The function will be called with `glob` as its only argument and must return a value of the right type for the option.
```
phated: the functions aren't called with the glob

lib/dest/index.js
```diff
@@ -12,25 +15,30 @@ function dest(outFolder, opt) {
     opt = {};
   }
 
+  var options = assign({}, opt, {
```
phated: this should just become `var sourcemaps = valueOrFunctions(['boolean', 'string', 'object'], opt.sourcemaps);`

lib/prepare-write.js
```diff
@@ -5,39 +5,34 @@ var path = require('path');
...
-  var options = assign({
-    cwd: process.cwd(),
-    mode: (file.stat ? file.stat.mode : null),
-    dirMode: null,
-    overwrite: true,
-  }, opt);
-  var overwrite = booleanOrFunc(options.overwrite, file);
-  options.flag = (overwrite ? 'w' : 'wx');
+  var defaultMode = file.stat ? file.stat.mode : null;
+  var options = assign({}, opt, {
```
phated: any reason for creating a new object here instead of just creating each property as it's own variable?

phated:
@hgwood just a few things.  Feel free to squash everything together if you want.

phated:
Oops, sorry for the accidental close.

lib/prepare-write.js
hgwood: Two reasons:
- it was done like that before I came in
- namespacing

Would you prefer individual vars suffixed with `Opt`? Same in `src/index.js`? Is that for performance reasons?

hgwood:
About the option name collision with through2 : what about naming the option `through2.read` (with a dot in the property name)?

```js
gulp.src(..., {
  read: true,
  'through2.read': function(...) {
    ...
  },
  sourcemaps: true,
  ...
}
```

[This code](https://github.com/gulpjs/vinyl-fs/pull/168/files#diff-2599ed3c945bd61f0161c2419a569f0bR35) would then become.

```js
// Don't pass `read` option on to through2
var read = options.read !== false;
options.read = options['through2.read'];
```

README.md
hgwood: Right.

lib/dest/index.js
hgwood: OK

phated:
I think there's still stuff to think about (like passing the `file` object to `.src` function options, `through2` options, etc) and docs cleanup + tests to do before this gets released, but I'm going to merge this in and we can work on that stuff separately.

Issue #165, opened 2016-04-25
The mkdirp module doesn't handle modes in the best way


phated:
* If a directory exists, the mode isn't updated - ref https://github.com/substack/node-mkdirp/issues/80
  * This could be solved by implementing our own mkdirp logic (in the vein of implementing our own `writeFile`)
* On OSX, the `fs.mkdir` call doesn't seem to set sticky, sgid or suid bits but `fs.chmod` does - ref https://github.com/gulpjs/vinyl-fs/blob/master/test/dest-modes.js#L339-L340
  * This would be solved by solving the first one because it would use chmod after the mkdir

PR #169, opened 2016-05-09
Custom mkdirp (Issue #165)

virtuakazib:
Custom mkdir implementation that include the ability to update the
permissions mode on a directory that already exists.

This is an initial cut at this feature that works with passing unit tests on OS X.  Haven't been able to verify Windows and Linux compatibility yet.

There are also probably coding convention and style preferences to fix.

### DIFF
--- a/lib/fileOperations.js
+++ b/lib/fileOperations.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var fs = require('graceful-fs');
+var path = require('path');
 var assign = require('object-assign');
 var isEqual = require('lodash.isequal');
 var isValidDate = require('vali-date');
@@ -9,6 +10,7 @@ var isValidDate = require('vali-date');
 // TODO include sticky/setuid/setgid, i.e. 7777?
 var MASK_MODE = parseInt('0777', 8);
 var DEFAULT_FILE_MODE = parseInt('0666', 8);
+var DEFAULT_DIR_MODE = parseInt('0777', 8);
 var APPEND_MODE_REGEXP = /a/;
 
 function closeFd(propagatedErr, fd, callback) {
@@ -161,7 +163,7 @@ function updateMetadata(fd, file, callback) {
   file descriptor after the write is complete.
   Most of the implementation taken from node core.
  */
-function writeFile(path, data, options, callback) {
+function writeFile(filepath, data, options, callback) {
   if (typeof options === 'function') {
     callback = options;
     options = {};
@@ -181,7 +183,7 @@ function writeFile(path, data, options, callback) {
   var flag = options.flag || 'w';
   var position = APPEND_MODE_REGEXP.test(flag) ? null : 0;
 
-  fs.open(path, flag, mode, onOpen);
+  fs.open(filepath, flag, mode, onOpen);
 
   function onOpen(err, fd) {
     if (err) {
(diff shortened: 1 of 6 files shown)


lib/fileOperations.js
```diff
@@ -196,11 +198,55 @@ function writeFile(path, data, options, callback) {
...
+  if (typeof mode === 'function') {
+    callback = mode;
+    mode = undefined;
+  }
+
+  var m = mode || DEFAULT_DIR_MODE;
+  var cb = callback || function() {};
+  dirpath = path.resolve(dirpath);
+
+  fs.mkdir(dirpath, m, function(er) {
```
phated: why does this attempt to use `fs.mkdir` before `mkdirp` (only if ENOENT)?

virtuakazib: The initial fs.mkdir call ends the recursion on success.  If a parent directory doesn't exist, it recursively calls 'mkdirp' up the path until fs.mkdir succeeds.  Then calls mkdir back down the path.

phated: It doesn't seem to call `mkdir` back down the path (cb is being called, ending the cycle).  This probably shouldn't be using `mkdirp` at all because it has to chmod each path segment

virtuakazib: The mkdir back down the path is on line 221.  It's actually another mkdirp call.

One clarifying question on mode:  
When mode is provided, all newly created directories will be created with that mode.  And currently, if the final directory on the path already exists, then chmod will be called to change the mode.

Do you want ALL existing directories in the provided dirpath parameter to change permission?

phated: Yeah, I believe the desire is to have every directory updated.

virtuakazib: >  This probably shouldn't be using mkdirp at all
This pull request removes the 'mkdirp' external dependency.  The only mkdirp is this function defined in fileOperation.js.

phated: Also, I was thinking it was still using the `mkdirp` module, totally forgot that the function was named mkdirp (ignore my comments about that).

virtuakazib: Gotcha.  I will try to add that mode detail in the next day or two.

erikkemperman:
@virtuakazib @phated I like this idea -- it removes the need to to subvert the setuid/setgid/sticky tests on MacOSX, as currently done [here](https://github.com/gulpjs/vinyl-fs/commit/3f9b9c19af09de9161739541a2b2eb866e1bf191#diff-35ff949480938a57085df3d1b0cf1512R340). Unfortunately I discovered too late (after [this PR](https://github.com/gulpjs/vinyl-fs/pull/162) was merged) what was actually up with that:

`mkdirp` (the module we were using, not the function here) restricts the mode bits to `777` (and excludes the bits set in `umask`) but the reason is likely that, on MacOSX, the resulting mode of creating a directory is unspecified if it includes "special" attributes. The manpages for the mkdir syscall says that in such cases we should follow up with a `chmod` which is effectively what this PR would do.

erikkemperman:
@virtuakazib @phated Any news on this? No pressure, I was just looking forward to not having to hack this `isDarwin` in the tests to make them pass on my Mac.

By the way, I was curious, why are the new tests in this PR skipped entirely for Windows? Wouldn't it be better to test the creation of dirs and only skip testing their modes?

phated:
@erikkemperman thanks for following up but it looks like changes are still to be made based on my comments.

phated:
@virtuakazib will you have anytime soon to make the changes?

phated:
This needs a few basic tests for Windows (that just check the directories are created).

Issue #158, opened 2016-02-26
Decouple opening and closing of file descriptors from src and dest


phated:
This could be done in a separate utility that might close descriptors lazily (having it ready in case a subsequent pipeline wants it).

Ref https://github.com/gulpjs/vinyl-fs/pull/151#issuecomment-183773006

erikkemperman:
@phated I've been looking into this a bit, thought I'd share:

- Implementing this will require considerable refactoring, especially of `src` which currently doesn't do FDs, at all;
- It will be tricky to get it right, I think, especially in combination with external tools or internal plugins interacting with the filesystem outside of vinyl-fs' control;
- I'm curious if this issue ought to be in the 3.0 milestone, the way I see this it would be a transparent change to end users (so long as it works, natch);
- Wondering if there will actually be a noticeable performance boost here, in common scenarios, that are not already benefiting from similar tricks in lower-level layers (caches).

I'm now thinking probably the last of these would be best to look at first, if that turns up a negative then I'd say we're done here. I'll not have a lot of time but would like to set up a simple experiment around this when I have the chance.

Thoughts?

erikkemperman:
BTW, in working up courage to attack this issue, I did some cosmetic stuff, for instance to make `src` and `dest` more (anti)symmetric in terms of things like module names. Might as well link to it here, if there's anything about it you like I can make a PR with a subset:
[commits](https://github.com/erikkemperman/vinyl-fs/commits/aesthetix)
[travis passes](https://travis-ci.org/erikkemperman/vinyl-fs/builds/135564452)

erikkemperman:
Forgot to mention: one little annoyance here is that in Node 0.10 we can't pre-open an FD if we want to use it to make readable/writable streams. That was only introduced in 0.12.

For some reason I have the impression that bumping minimum require node version is not really an option, have I got that right? If so, that'd mean an extra little complication for this issue.

phated:
* I'm excited about the refactoring, it is something I've wanted to happen since the `.dest()` stuff was being worked on.
* I totally understand that we might not want to hold them open due to people interacting (against our guidelines) with the filesystem.
* The reason I wanted this in the 3.0 milestone was just out of safety.  I'm scared of any sweeping changes to land in 2.x after what happened with `utimes`
* I actually have no idea if there will be a performance boost (and I don't even know if there was a performance decrease when `utimes` was added).  I'm interested in vinyl-fs being as fast as possible but I don't see it causing problems right now.

[...]

PR #173, opened 2016-06-07
Remove superfluous argument from write-xxx modules

erikkemperman:
As discussed [here](https://github.com/gulpjs/vinyl-fs/issues/158).


phated:
Marking this as "Breaking" (even though it might not be) because someone might be mutating the file.path while we are using it.  Not sure if that is possible or actually done anywhere, but just staying on the safe side.

PR #182, opened 2016-06-29
Consistent callback names

erikkemperman:
@phated This was in my aesthetix branch I linked to earlier, but couldn't make a PR for it at the time because of conflicts. It is completely cosmetic, but you said you liked it, so here goes.


phated:
Thanks for all these PRs!

erikkemperman:
NP, glad something came out of that decouple-fd issue -- which I agree is probably more trouble than it's worth, I noticed you closed it. Although for symmetry I guess `src` might also hand off the fs interaction to `file-operations`?

PR #187, opened 2016-07-02
Add relative symlink support

perrin4869:
Adds support for creating symlinks which point to a file relatively

### DIFF
--- a/lib/symlink/index.js
+++ b/lib/symlink/index.js
@@ -2,9 +2,20 @@
 
 var through2 = require('through2');
 var fs = require('graceful-fs');
+var path = require('path');
+var valueOrFunction = require('value-or-function');
+
 var prepareWrite = require('../prepare-write');
+var defaultValue = require('../default-value');
+
+var boolean = valueOrFunction.boolean;
 
 function symlink(outFolder, opt) {
+  if (!opt) {
+    opt = {};
+  }
+
+  var rel = defaultValue(false, boolean(opt.relative));
   function linkFile(file, enc, cb) {
     var srcPath = file.path;
     var symType = (file.isDirectory() ? 'dir' : 'file');
@@ -12,6 +23,14 @@ function symlink(outFolder, opt) {
       if (err) {
         return cb(err);
       }
+
+      if (rel) {
+        var dstDir = (fs.existsSync(file.path) && symType === 'dir')
+          ? file.path
+          : file.dirname;
+        srcPath = path.relative(dstDir, srcPath);
+      }
+
       fs.symlink(srcPath, file.path, symType, function(err) {
         if (err && err.code !== 'EEXIST') {
           return cb(err);
--- a/lib/symlink/index.js
+++ b/lib/symlink/index.js
@@ -15,8 +15,8 @@ function symlink(outFolder, opt) {
     opt = {};
   }
 
-  var rel = defaultValue(false, boolean(opt.relative));
   function linkFile(file, enc, cb) {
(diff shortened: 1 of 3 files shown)


lib/symlink/index.js
```diff
@@ -2,16 +2,35 @@
...
   function linkFile(file, enc, cb) {
     var srcPath = file.path;
     var symType = (file.isDirectory() ? 'dir' : 'file');
     prepareWrite(outFolder, file, opt, function(err) {
       if (err) {
         return cb(err);
       }
+
+      if (rel) {
+        var dstDir = (fs.existsSync(file.path) && symType === 'dir')
```
phated: Probably shouldn't use `existsSync` here.  It is deprecated, a sync method and I think we just assume the files exist with everything else.

lib/symlink/index.js
```diff
@@ -2,16 +2,35 @@
...
+          ? file.path
+          : file.dirname;
+        srcPath = path.relative(dstDir, srcPath);
```
phated: I think vinyl has a `file.relative` property that we could use instead of this branch; If it does, you'd be able to just use a ternary when initializing `srcPath` on line 20

test/symlink.js
```diff
@@ -183,44 +184,54 @@ describe('symlink stream', function() {
...
-      cwd: __dirname,
-      path: inputPath,
-      contents: expectedContents,
-      stat: {
-        mode: expectedMode,
-      },
-    });
-
-    var buffered = [];
+  [{
```
phated: I prefer not to have generative testing.  It allows for separate refactoring of tests

test/symlink.js
```diff
@@ -268,46 +279,56 @@ describe('symlink stream', function() {
...
-
-    var expectedFile = new File({
-      base: inputBase,
-      cwd: __dirname,
-      path: inputPath,
-      contents: null,
-      stat: {
-        isDirectory: function() {
-          return true;
+  [{
```
phated: I prefer not to have generative testing.  It allows for separate refactoring of tests

phated:
@perrin4869 awesome work.  I just had a few comments/suggestions

lib/symlink/index.js
```diff
@@ -2,16 +2,35 @@
...
+var defaultValue = require('../default-value');
+
+var boolean = valueOrFunction.boolean;
 
 function symlink(outFolder, opt) {
+  if (!opt) {
+    opt = {};
+  }
+
+  var rel = defaultValue(false, boolean(opt.relative));
```
phated: this should be inside the `linkFile` function and `boolean` should be passed `file` as the 2nd argument.

lib/symlink/index.js
perrin4869: `file.relative` returns the difference between `file.path` and `file.base`, which is not quite what we need here. We need the relative path between the target of the symlink and the path where the symlink lives

perrin4869:
pushed some improvements to the implementation!
There is a bit of code duplication with `prepareWrite`. The code that calculates the output directory I took mostly from there. Might be worth to look into factoring it out into helper functions.

mfeltscher:
This is a very useful feature - love it! Are there any plans to release a new version of `vinyl-fs` including this feature?

phated:
It will land in 3.0 due to other breaking changes.

mfeltscher:
Sounds good. Is there an ETA for 3.0?

phated:
As soon as the outstanding issues are resolved. We all do this in our free time. Feel free to submit a PR if you can.

PR #188, opened 2016-07-05
Added support for changing uid/gid (fix #157)

### LINKED ISSUE #157: Support changing uid/gid from the vinyl object to the file on disk
We now allow `mtime`, `atime` and `mode` to be set on the actual file if they exist on the vinyl object.  We should also support `uid` and `gid` through use of `fs.fchown`.

Ref https://github.com/gulpjs/vinyl-fs/pull/151#issuecomment-183773006

stephanvl:
This PR adds support for changing the uid/gid of a file/directory.

Changing the uid/gid of a file is only possible as the root user and will fail silently if this run as a non root user.

What do you think we should do in this case?

### DIFF
--- a/lib/file-operations.js
+++ b/lib/file-operations.js
@@ -26,6 +26,18 @@ function closeFd(propagatedErr, fd, callback) {
   }
 }
 
+function isValidUnixId(id) {
+  if (typeof id !== 'number') {
+    return false;
+  }
+
+  if (id < 0) {
+    return false;
+  }
+
+  return true;
+}
+
 function getModeDiff(fsMode, vinylMode) {
   var modeDiff = 0;
 
@@ -67,33 +79,24 @@ function getTimesDiff(fsStat, vinylStat) {
 }
 
 function getOwnerDiff(fsStat, vinylStat) {
-  if (typeof vinylStat.uid !== 'number' &&
-      typeof vinylStat.gid !== 'number') {
+  if (!isValidUnixId(vinylStat.uid) &&
+      !isValidUnixId(vinylStat.gid)) {
     return;
   }
 
-  var uid = fsStat.uid; // Default to current uid.
-  if (typeof vinylStat.uid === 'number' &&
-      vinylStat.uid >= 0) {
+  var uid = isValidUnixId(fsStat.uid) ? fsStat.uid : undefined; // Default to current uid.
+  if (isValidUnixId(vinylStat.uid)) {
     uid = vinylStat.uid;
   }
 
-  var gid = fsStat.gid; // Default to current gid.
-  if (typeof vinylStat.gid === 'number' &&
-      vinylStat.gid >= 0) {
+  var gid = isValidUnixId(fsStat.gid) ? fsStat.gid : undefined; // Default to current gid.
+  if (isValidUnixId(vinylStat.gid)) {
     gid = vinylStat.gid;
   }
 
   if (isEqual(uid, fsStat.uid) &&
       isEqual(gid, fsStat.gid)) {
     return;
-  }
-
-  if (uid < 0) {
-    uid = undefined;
-  }
-  if (gid < 0) {
-    gid = undefined;
   }
 
   var ownerDiff = {
@@ -197,13 +200,13 @@ function updateMetadata(fd, file, callback) {
           file.stat.mtime = timesDiff.mtime;
         }
         if (ownerDiff) {
-          owner(fchmodErr, futimesErr);
+          return owner(fchmodErr || futimesErr);
         }
         callback(fchmodErr || futimesErr);
       }
     }
 
-    function owner(fchmodErr, futimesErr) {
+    function owner(earlierErr) {
       fs.fchown(fd, ownerDiff.uid, ownerDiff.gid, onFchown);
 
       function onFchown(fchownErr) {
@@ -211,7 +214,7 @@ function updateMetadata(fd, file, callback) {
           file.stat.uid = ownerDiff.uid;
           file.stat.gid = ownerDiff.gid;
         }
-        callback(fchmodErr || futimesErr || fchownErr);
+        callback(earlierErr || fchownErr);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
   return timesDiff;
 }
 
+function getOwnerDiff(fsStat, vinylStat) {
+  if (/^win/.test(process.platform)) {
```
phated: there should be no reason to do this.  In `updateMetadata`, we exit early on windows (it doesn't have `process.getuid()`)

lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
...
+    return; // Always return undefined on windows.
+  }
+
+  if (typeof vinylStat.uid !== 'number' &&
```
phated: this should probably be abstracted to a method called `isValidUid()` incase we need to add extra logic and it keeps it consistent with `getTimesDiff`

lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
...
+      typeof vinylStat.gid !== 'number') {
```
phated: this should probably be abstracted to a method called `isValidGid()` incase we need to add extra logic and it keeps it consistent with `getTimesDiff`

lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
...
+    return;
+  }
+
+  var uid = fsStat.uid; // Default to current uid.
+  if (typeof vinylStat.uid === 'number' &&
+      vinylStat.uid >= 0) {
```
phated: this can go in the `isValidUid` method

lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
...
+    uid = vinylStat.uid;
+  }
+
+  var gid = fsStat.gid; // Default to current gid.
+  if (typeof vinylStat.gid === 'number' &&
+      vinylStat.gid >= 0) {
```
phated: this can go in the `isValidGid` method.  If these are exactly the same, maybe there's a better name to give them and use it for both.

lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
...
+    gid = vinylStat.gid;
+  }
+
+  if (isEqual(uid, fsStat.uid) &&
+      isEqual(gid, fsStat.gid)) {
+    return;
+  }
+
+  if (uid < 0) {
+    uid = undefined;
```
phated: this should be defaulted when `isValid...` is checked

lib/file-operations.js
```diff
@@ -66,6 +66,48 @@ function getTimesDiff(fsStat, vinylStat) {
...
+  }
+  if (gid < 0) {
+    gid = undefined;
```
phated: this should be defaulted when `isValid...` is checked

lib/file-operations.js
```diff
@@ -149,9 +200,24 @@ function updateMetadata(fd, file, callback) {
           file.stat.atime = timesDiff.atime;
           file.stat.mtime = timesDiff.mtime;
         }
+        if (ownerDiff) {
+          owner(fchmodErr, futimesErr);
```
phated: this should probably be `||` like the other callback but i'm not sure

phated: missing the `return` here so the `callback` below would always be called at the wrong time

test/file-operations.js
```diff
@@ -333,6 +334,221 @@ describe('getTimesDiff', function() {
...
+      gid: 1001,
+    };
+    expected = {
+      uid: 1000,
+      gid: 1001,
+    };
+
+    var result = getOwnerDiff(fsStat, vfsStat);
+
+    expect(result).toEqual(expected);
```
phated: this is failing on appveyor.

test/file-operations.js
```diff
@@ -333,6 +334,221 @@ describe('getTimesDiff', function() {
...
+    };
+    var vfsStat = {
+      uid: 1001,
+      gid: 1000,
+    };
+    var expected = {
+      uid: 1001,
+      gid: 1000,
+    };
+    if (isWindows) {
```
phated: If Windows doesn't provide a `uid` or `gid` when `fs.stat` is called, we might as well skip all these tests.  I will have to check this later.

phated:
@stephanvl very nice work.  I just had some suggestions and the appveyor CI is still failing.  Let me know once you've had a chance to clean it up.

stephanvl:
@phated I removed the latter 2 commits which added the windows specific code, added a `isValidUnixId` method to check the validity of the uid and gid and I added the return statement to fix the callback issue.

I also updated the `owner` method to expect one error instead of separate errors for the owner and times.

test/file-operations.js
```diff
@@ -333,6 +362,206 @@ describe('getTimesDiff', function() {
...
+      uid: 1000,
+      gid: undefined,
+    };
+    var vfsStat = {
+      uid: 1001,
+      gid: undefined,
+    };
+    var expected = {
+      uid: 1001,
+      gid: undefined,
```
phated: fs.fchown throws if uid or gid is undefined.  Do we need to handle this by returning `undefined` if we don't have a complete diff?

phated:
@stephanvl this looks near perfect.  Just 1 thing that I noticed while poking around in the node repl.  Commented inline.

stephanvl:
@phated Good catch, I will change it to return undefined when either uid or gid is undefined as there is no way of knowing what it should be at that point.

Then there is only the question what we should do about the fact that chown only works when the user is root (fail silently, throw an error or log a message and continue)?

phated:
@stephanvl we already handle that with our `isOwner` check inside the `updateMetadata` function.  It checks the user owns the file or is effectively root.

stephanvl:
@phated but you must be root for chown to work, being the owner of a file is not enough.

https://en.wikipedia.org/wiki/Chown

phated:
Those errors should be bubbled up, I believe.

stephanvl:
graceful-fs gracefully ignores EPERM errors, which changing the owner as a non root user is

phated:
👍  I'm fine with that then

phated:
@stephanvl amazing work!  Thanks a ton

PR #194, opened 2016-07-22
(Don't Merge) Test refactor

phated:
Opening this for review.  Once I have some feedback, I'll push it to master to keep the 2 commits separate.

cc'ing anyone that might be remotely familiar since there's a lot of changes.

@contra @terinjokes @erikkemperman @jonschlinkert


contra:
LGTM

terinjokes:
LGTM

lib/prepare-write.js
```diff
@@ -44,7 +44,9 @@ function prepareWrite(outFolder, file, opt, callback) {
   file.stat.mode = options.mode;
   file.flag = options.flag;
   file.cwd = cwd;
-  file.base = basePath;
+  // Ensure the base always ends with a separator
+  // TODO: add a test for this
+  file.base = path.normalize(basePath + path.sep);
```
jonschlinkert: since `path.normalize()` will fix `'foo//bar'` to `foo/bar` (or backslashes if windows), to be safe I would consider doing `basePath + '/' + path.sep`. Or just `path.join()`, but if you're doing normalize join isn't really needed.

jonschlinkert:
this looks great

erikkemperman:
@phated I sent a PR to this branch with an utterly pedantic change. So yeah, this looks awesome 👍

lib/prepare-write.js
phated: I actually think this needs to be pushed down into vinyl but I will do that separately from this PR

PR #196, opened 2016-07-23
Add space to not-owned notices

erikkemperman:
Pedantic, I know, but it looked off :-)

### DIFF
--- a/test/not-owned.js
+++ b/test/not-owned.js
@@ -32,7 +32,7 @@ describe('.dest() on not owned files', function() {
   it('does not error if mtime is different', function(done) {
     if (dirStats.uid !== 0 || fileStats.uid !== 0) {
       console.log('Test files not owned by root.');
-      console.log('Please chown ' + notOwnedBase + ' and' + notOwnedPath + ' and try again.');
+      console.log('Please chown ' + notOwnedBase + ' and ' + notOwnedPath + ' and try again.');
       this.skip();
       return;
     }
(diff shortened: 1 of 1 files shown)


phated:
Good catch

erikkemperman:
Huh, I just chowned this dir and file (to root) and now I have failing tests. Need to look further into this, but that's not supposed to happen, right?

phated:
You have to make sure the files have the correct permissions too.  I had to chmod and chown to get them to work locally.  They work on travis so I'm not too concerned.

phated:
They are working on my machine also.

erikkemperman:
Yeah, my git sets not-owned.txt up with `644` (-rw-r--r--) after clone and the file needs to be `666` for the tests to work (with the user having followed the notice and chowned these fixtures). 

Interestingly though, git doesn't pick up a change if I chmod this file to `666`. I thought git accounts for modes (as opposed to uid/gid). In fact I made a similar change to a file over in the `gulp` repo recently, which was executable for no reason...

But here it doesn't work. I played around a little, and it seems git only cares about the `x` privilege to the current user, it ignores `r` and `w` for this user, and it ignores all of them for "others" and "group".

That it does work on Travis (and I presume Linux generally) but not on Mac makes me think the (default?) behaviour of Git is distinct in this regard. I have to call it day now but would suggest having the test check if the file has `w` privilege for "group" (if current user is part of the group) and/or "others" and append the need to chmod to the printed notice where applicable?

erikkemperman:
@phated I just quickly followed this up and added a notice about chmod to the not-owned test. It's more pedantry, I guess, sorry about that. I trust you'll let me know if it gets to be annoying...

https://github.com/erikkemperman/vinyl-fs/commit/639215b869d30d418eec59786740abe6d67a3234

erikkemperman:
PS I may be missing something obvious, but why does the directory have to be chowned as well? The chown to root requires sudo anyway, so access to the directory should not be a problem.

phated:
It probably doesn't need to be. I just did it to be safe.

erikkemperman:
All right, I see. I didn't want to presume and make a PR of the chmod notice, but it's there if you want it.

phated:
Feel free to send a PR.  Easier for me to merge.

Issue #202, opened 2016-08-12
custom WriteStream on writes end of stream


doowb:
When a stream is large enough, only the end is written to the file.

The code [here](https://github.com/gulpjs/vinyl-fs/blob/master/lib/file-operations.js#L401) is using `this.pos` which is set to `null` when the append flag is set. This is good for `writeFile` which writes the entire file at once, but it should be different inside the `WriteStream`.

I've tried just passing in `null` so the contents will be written to the end of the file and also, updating the `this.pos` in the `onWrite` callback. I think the latter is a better approach since the append flag may still be used.

I think updating [the onWrite callback](https://github.com/gulpjs/vinyl-fs/blob/master/lib/file-operations.js#L407-L415) would look like this:

```js
  function onWrite(writeErr) {
    if (writeErr) {
      self.destroy();
      callback(writeErr);
      return;
    }

    if (self.pos !== null) {
      self.pos += data.length;
    }
    callback();
  }
```

I can submit a PR if you'd like. I'm not sure the best approach for testing this.

Thanks to @criticalmash for discovering this and working with me on debugging.

phated:
@doowb that's interesting because I copied most of the logic from node's write stream. From my initial digging, I found that fs.write maintains the position internally if you don't pass a position. Can you provide an example repo of this?

doowb:
`fs.write` does maintain the position internally and uses that if you pass a non number in. But [this line](https://github.com/gulpjs/vinyl-fs/blob/master/lib/file-operations.js#L401) is passing in `this.pos`, which is set to `0` [here](https://github.com/gulpjs/vinyl-fs/blob/master/lib/file-operations.js#L352).

I'll put a repo together showing it so we can get a test out of it.

phated:
Oh, that might have happened during refactoring (notice the double semi-colon).  We can probably get away with hardcoding that as `null`.

doowb:
That was my first thought too, but will that just "append" to the file if it already has contents?

phated:
@doowb the `fs.open` call is still made with the proper flag, so I would hope so.  Otherwise something is seriously messed up in node core.

doowb:
> fs.open call is still made with the proper flag

K, that's the part I missed. Passing `null` should be good.

doowb:
Here's a simple repo showing the issue in case it can be used for a test: https://github.com/doowb/vinyl-fs-write-stream

criticalmash:
[...]

PR #203, opened 2016-08-12
Custom write stream fix

doowb:
This is a fix and tests for #202.

I added a large file to the fixtures that demonstrates the issue.
I added tests to the file operations test file and the dest test file. These needed to use `fs.createReadStream` directly since `from` was writing the entire string to the stream at once.


phated:
Awesome, you can actually use `from` with a function to generate a really big file instead of including fixtures.

doowb:
@phated I updated the tests to use `from` to create the read stream.

phated:
Thanks for all the work on this and finding this bug!  I'm going to merge and then tweak a few things.

Issue #210, opened 2016-09-12
Symlinks on Modern Windows with latest node/npm.


afinnell:
Theree are not only two options to creating a symlink. Either Node should change it's default behavior for the symlink type 'dir', or module makes should start using 'junction' instead of 'dir' when on windows. The symlink 'dir' on Windows does not behave the same way as a POSIX style symlink. 

The proper symlink for windows is a Junction which makes more sense when comparing it to a *nix style symlink. 

symlink's of type 'Dir' cannot be created on windows without a Run As verb which is completely unacceptable. I've patched dep-linker to work on all platforms and not require the process to be ran as an actual Administrator. 

This is the equivalent of REQUIRING a Linux user to run gulp as root. Not sudo'ing the process. But literally requiring them to run gulp as root user. I cannot imagine that would ever make it into a pull request, why does it make it into pull requests on the Windows side?

contra:
@afinnell I don't think any of us use windows so oversights like this happen - no need to get angry at us. We try our best to support windows even though 99% of our users are on unix systems, but that does put some responsibility on the 1% who do use windows to report issues - so thank you for doing that.

If I'm understanding you right, you're saying [this](https://github.com/gulpjs/vinyl-fs/blob/master/lib/symlink/index.js#L22) should be changed to use `junction` for directories on windows, right?

afinnell:
@contra My apologies. That Issue came across more hostile then intended. There's a bit of a closd-circle regarding npm's, fixes, and the ability to easily get them back onto a developers system using NPM. Meaning it's difficult to path modules ourselves and then use them across all machines. We have to wait for the author's to merge then release. 

And it is also not so easy to create our own modules without going through the npmjs processs. At any rate I've been on the hunt to find something to easily symlink node_modules to a /lib dir and each symlink module always uses 'dir'. 

At any rate I've submitted a pull request at https://github.com/gulpjs/vinyl-fs/pull/211

On Windows, the reason the DIR symlink requires such a heavy lock down is because it follows symlinks based on the LOCAL USER's relative path. So accessing symlinks across a shared drive can result in a User's own local directory getting resolved instead of the intended original link. Personally I think it's Microsofts fault for renaming it to Junction's. 

But, in short, yes, what I have found is that Windows Junction === Linux Directory symlinks. If I've overlooked anything it would be very helpful for someone to point it out before I go too far down the Junction path. 

Thanks!

afinnell:
[...]

PR #211, opened 2016-09-12
#210 Fix Windows symbolic links requiring admin

afinnell:
Default to junctions instead of dir on windows unless told otherwise.


contra:
Tests are failing ^

phated:
@afinnell any progress on getting the tests to pass?

afinnell:
My apologies. This notification became swept up in the wrong folder. I'll take a look.

afinnell:
@phated I believe I've fixed the errors now that I see what they were. I feel slightly silly but it doesn't appear that GitHub has an easy way to update an existing Pull Request with an additional commit. 

Should I rebase the commits in this branch and push it back or will that break the CI server since it's already cloned the repo and checked out the branch? I can close the PR and create a new one if that's an easier solution.

phated:
You can just push the commit and it will be added to the PR. Rebase also works

afinnell:
So I don't think Test "symlink stream can create relative links for directories:" is going to work with the junctions. I'll change the logic around to support junctions instead of them being the default. This way the tests stay the same and no one is impacted. however, I still don't understand how these tests even pass unless they are ran as the Administrator or as a User that is not in the Administrators group. 

I'm beginning to think that running as a User who is in the Administrator group is not the way other developers run on Windows. Not running as Administrator User, but as a User in the Administrator group. The OS Kernel hard stops ANY User that is in the Administrator group from creating symlinks and there is no setting I can find to change this. Now if I create a dummy account that is not in the Administrators group then I can create symlinks. This is all backwards in my mind and wouldn't mind someone explaining this logic to me. 

I'll post the update shortly.

afinnell:
Here is some irony for anyone following the ticket:

If one were to use Bash for Windows, and create a symbolic link using the `ln -s` command in a Bash shell, it shows up as a JUNCTION on Windows no matter what. Even if it's a symbolic link to a file. Which obviously doesn't work.

phated:
Seems like a linting error crept in.  Please fix.

afinnell:
Fixed lint errors again. Not sure how those spaces keep ending up on the end. Toughest lint rules I've ever seen, but makes for nice clean code. 

I also added a hard-link stream target and the mocha tests to follow.

afinnell:
@phated @contra If everyone is comfortable with these changes (which should not break existing functionality / gulps) can we get this into the next release, asap? NPM doesn't make it easy to use your own version of a build. I realize it's possible, it's just not easy especially when you have a team. 

If this can't make it into the build that is not a problem, I just need to know so that I can create my own npm gulp module. 

My appreciation!

Andrew

contra:
@afinnel you can put `"vinyl-fs": "afinnell/vinyl-fs"` in your package.json and it'll use your version

LGTM, but missing documentation

afinnell:
@contra The wiki pages or the .MD files? Or both?

contra:
@afinnel The README. Wasn't aware we had wiki pages (?), I thought i had disabled that on every repo since they are typically worse than just markdown files.

contra:
@afinnell Correction: we don't have a wiki

image

phated:
@afinnell I still have to review this but i wanted to let you know it will go in the 3.0 release because we have many breaking changes in the master branch and rebasing this into a different release would be a headache.  We still have a few things to get done before the 3.0 release.

afinnell:
@phated Not a problem. After feeling ignorant and not realizing I could npm install afinnell/vinyl-fs I'm set. I can merge these changes into the 3.x mainline if you need me to. Even if it requires a rework, I can redo it once 3.x is stable. Thanks!

phated:
I reviewed this and it looks good. I need to figure out exactly when I'm going to pull it in.  Thanks for all the work @afinnell

phated:
@afinnell actually, can you squash the commits and rebase/cherry-pick onto master?  I've just done some rebasing to cleanup commits and it seems to have diverged your PR.  Thanks again.

afinnell:
@phated Not a problem, it'll be my pleasure.

K05:
@afinnell @phated wondering what is the status w/ this fix?

phated:
Not even reviewed yet; there's too much else to work on. Looks like it still needs to be rebased and squashed, too.

phated:
@afinnell I just rebased this and started to review.  It seems the wrong thing is being tested.  You added tests for `.src` reading the hardlinks but nothing where your new `.link` method is being used.

afinnell:
I will look at my local repo and ensure the latest commits were pushed. Been out of hand and thus unresponsive. I appreciate the heads up on this.

phated:
Solved by 5ae7d7e4c674cdd6bafd8d29eb332c0adbcb760b

PR #219, opened 2017-01-24
Fix : repository url error

davidhuang1550:
Hi, after reading your package.json file I have noticed that the repository URL is invalid. I have made the following changes, please look over them.


phated:
It's valid by npm standards.

PR #225, opened 2017-03-28
chore(followSymlinks): Rename to resolveSymlinks

### LINKED ISSUE #205: Revisit the naming of `followSymlinks` option
This was brought up in https://github.com/gulpjs/gulp/issues/1741

Basically, `vinyl-fs` has an option called `followSymlinks` that reads the contents from the linked file and creates a "hard file" of that instead of a symlink.  If you set the option to `false`, we output symlinks in `dest()`.  This is unfortunately named because it implies we traverse an entire tree of symlinks, which isn't the case because node-glob has a (similarly named) option called `follow` which follows symlinks down a tree; they default this option to `false` because circular symlinks can cause problems.

Our `followSymlinks` option doesn't properly describe what is happening and should probably be renamed in the next breaking release.

@erikkemperman something to keep in mind if you work on the `src` options.

kyleholzinger:
Not sure if I understood the issue, but this is an attempt to resolve #205 (assuming `resolveSymlinks` was the name that was landed on 😄  )

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Type: `Boolean`
 
 Default: `false`
 
-##### `options.followSymlinks`
+##### `options.resolveSymlinks`
 
 Whether or not to recursively resolve symlinks to their targets. Setting to `false` to preserve them as symlinks and make `file.symlink` equal the original symlink's target path.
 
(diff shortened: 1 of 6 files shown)


phated:
@kyleholzinger thanks for sending this.  I just noticed that this was completed while I cherry-picked @erikkemperman's commits.  Sorry for not getting yours merged.

PR #228, opened 2017-04-18
Use ctime for since filter

### LINKED ISSUE #226: Use ctime for since filter?
Hi!

When `gulp.src` searches for files with `since` option, vinyl-fs compares mtime of each file with `since` value ([here](https://github.com/gulpjs/vinyl-fs/blob/master/lib/filter-since.js#L14)). But what if a file wasn't changed? 

When I just paste file to watched directory, its mtime does not change. And `gulp.src` doesn't see this file. But ctime changes. May be `filterSince` should compare ctime too?

igoradamenko:
This PR fix #226.

I did some research and write a test for OS (macOS / Windows) copy-paste behaviour. So when user copy-paste file, OS doesn't change mtime. And the new test cover this case (as much as possible). 

But it must be noted, that @phated was partially right about ctime and file reading. If Windows user create file (e.g. using context menu) and then open it first time, ctime changes twice: after creation and after opening. But it works just one time. Look as these screenshots:

screen shot 2017-04-19 at 00 48 35

Above you see three fs.stat results. The first was printed after file creation, the second — after file opening, the third — after tens times of opening (using Notepad and Wordpad, not just one editor). 

Next, I copy-pasted this file, then printed stat, then opened, then printed stat again:

screen shot 2017-04-19 at 00 49 16

Note, that mtime of copy is the same as mtime of original file! Behaviour of ctime is the same also. So, when I opened the copy again and again (after first opening), ctime didn't changed (the latest stat):

screen shot 2017-04-19 at 00 50 18

I suppose that text editors on Windows change file metadata or something else on first open (or may be some rights, who knows). I don't sure that this problem is more important that copy-paste issue. So I think vinyl-fs should use ctime in `since` 🐨


phated:
@igoradamenko thanks.  I'm going to hold-off on reviewing this while the big 3.0 changes are in progress

phated:
also cc @erikkemperman since he worked on a bunch of this stuff

erikkemperman:
I think the idea is that mtime reflects the last content change and ctime the last metadata (inode) change.

It might make sense to include the latter, but I don't think we should ditch the former. Maybe compare with `max(ctime, mtime)` to be sure?

Although having said that I'm not sure if it's even possible to end up with `mtime > ctime`, I guess it depends on the tools used to modify the file. I suppose truncating or appending to an existing file might do that?

Also, is node 0.10 still supported? The docs say that the semantics of ctime on windows changed in 0.12.

igoradamenko:
> Although having said that I'm not sure if it's even possible to end up with mtime > ctime

It seems like not possible, 'cause update of mtime also changes ctime.

> Also, is node 0.10 still supported? The docs say that the semantics of ctime on windows changed in 0.12.

Wow, I really didn't think about it. Yes, ctime in node < 0.12 contains creation time (birthtime) of file in Windows:

> Prior to Node v0.12, the ctime held the birthtime on Windows systems. Note that as of v0.12, ctime is not "creation time", and on Unix systems, it never was.

@phated, will v3.0 support node 0.10? If it will, so we can use something like this:

```js
var timestamp = process.version < 'v0.12' ? 'mtime' : 'ctime';
return file.stat && file.stat[timestamp] > date;
```

But I'm not sure that it's a good solution — behaviour depends on node version.

erikkemperman:
You seem to be right, changing ctime implies changing mtime, but not vice versa. So forget about my `max` idea, it doesn't apply.

But do we actually want to pass a file that's copied in if it hasn't changed (content) since the last build? I think I would argue we *don't*, in most cases, because the idea is to not re-build the exact same artifacts over and over -- and this is a function of content, not metadata.

For what it's worth, GNU make uses mtime, not ctime:
http://git.savannah.gnu.org/cgit/make.git/tree/file.c

If you want to force it, you can always `touch` the file -- or, you know, simply not use the `since` option?

If we want to use ctime instead of mtime after all, there are a few more places you'll want to look at, such as https://github.com/gulpjs/vinyl-fs/blob/master/lib/file-operations.js

It looks like Node 0.10 is still supported: https://github.com/gulpjs/vinyl-fs/blob/master/.travis.yml#L8 and I seem to remember @phated wanting to keep it around for the time being?

igoradamenko:
> But do we actually want to pass a file that's copied in if it hasn't changed (content) since the last build?

Yes. Look at the #226. If we copy-paste any file that created / modified _before_ we run gulp, it won't be detected. For example, I have an images in folder A and project in folder B. I run build in B, then copy-paste images from A to B and gulp / vinyl-fs doesn't see them.

Maybe it's possible to detect new files using another way, but I don't know how, 'cause vinyl-fs doesn't watch files (AFAIK).

> For what it's worth, GNU make uses mtime, not ctime: http://git.savannah.gnu.org/cgit/make.git/tree/file.c

Sorry, I didn't dive in source of `make`, but it has such staff:

```c
if (f->last_mtime == UNKNOWN_MTIME)
    puts (_("#  Modification time never checked."));
else if (f->last_mtime == NONEXISTENT_MTIME)
    puts (_("#  File does not exist."));
else if (f->last_mtime == OLD_MTIME)
    puts (_("#  File is very old."));
```

So I think last_mtime it's a field of file object in `make` memory, not a file mtime. And `make` mark all new files with `last_mtime: UNKNOWN_MTIME`, and then compare this mark with file's real mtime and change it. `make` use such way to make possible to detect new files.

> or, you know, simply not use the since option?

Yeah, I'm using [gulp-cached](https://github.com/contra/gulp-cached) for that, but it's sad, you know. gulp (vinyl-fs) has a simple tool for file changes detection that sometimes does not work 🙁 

> If we want to use ctime instead of mtime after all, there are a few more places you'll want to look at, such as https://github.com/gulpjs/vinyl-fs/blob/master/lib/file-operations.js

As I see, `getTimesDiff` function and others compare not just mtime, but atime too. So atime changes always together with ctime / mtime, and there is nothing to change.

> It looks like Node 0.10 is still supported

I do not insist on this solution and don't require to merge this PR. Honestly, I don't completely sure that it's absolutely right to use ctime instead of mtime. I just trying to say: “Hey, guys, maybe there is another solution that can make `since` filter a little bit better”. But if it's not so.. ¯\\\_(ツ)_/¯ just don't merge it 🐨 

From the clash of opinions emerges the truth and so on :)

erikkemperman:
OK, sorry, I linked the wrong file: http://git.savannah.gnu.org/cgit/make.git/tree/filedef.h. It references `st_mtime` from the stat struct.

But the important thing is the desired behaviour, of course. So, I am not against this PR or anything. I was just wondering if it is expected that if we replace a file by a new file with the same contents (copy/paste) that it passes the filter. I would expect it not to, but obviously you do :-)

test/src.js
```diff
@@ -299,6 +299,33 @@ describe('.src()', function() {
     ], done);
   });
 
+  it('streams a file changed after since (when file copy-pasted)', function(done) {
+    var testFilePath = inputPath + 'copy-paste-test';
+    var stat = fs.statSync(inputPath);
+    var fd, lastUpdateDate;
+
+    // When in OS user copy-paste file, its mtime does not changed
```
phated: Instead of emulating this, can you use `child_process` to copy a file?  You can create separate tests for each OS

phated:
This needs to be rebased and I'd like to see some tests that aren't emulate that demonstrate the problem.

test/src.js
igoradamenko: Okay, I will try to use it.

igoradamenko:
@phated, can you explain what do you mean about `child_process`? exec / execSync with something like `cp src dest`, right?

If so, there is a problem. OS's copy-paste and `cp` work differently. There is the PoC:

Create file in terminal using `touch` and get its ctime & mtime:
```bash
$ touch test.txt
$ stat test.txt
  File: "test.txt"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (  501/   dante)  Gid: (   20/   staff)
Device: 1,4   Inode: 27488513    Links: 1
Access: Mon Jul 24 21:11:53 2017
Modify: Mon Jul 24 21:11:53 2017 # !
Change: Mon Jul 24 21:11:53 2017 # !
```

Copy this file to another using `cp` and get ctime & mtime again:

```bash
$ cp test.txt test2.txt
$ stat -x test2.txt
  File: "test2.txt"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (  501/   dante)  Gid: (   20/   staff)
Device: 1,4   Inode: 27488517    Links: 1
Access: Mon Jul 24 21:12:08 2017
Modify: Mon Jul 24 21:12:08 2017 # !
Change: Mon Jul 24 21:12:08 2017 # !
```

So, `cp` changes ctime & mtime, right? Okay, let's copy-paste `test2.txt` using OS. In my El Capitan I've opened folder with file, pressed ⌘C on file and pressed ⌘V. OS created file `test2 copy.txt`. Let's look at its ctime and mtime:

```bash
$ stat -x test2\ copy.txt
  File: "test2 copy.txt"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (  501/   dante)  Gid: (   20/   staff)
Device: 1,4   Inode: 27488532    Links: 1
Access: Mon Jul 24 21:12:33 2017
Modify: Mon Jul 24 21:12:08 2017 # !
Change: Mon Jul 24 21:12:33 2017 # !
```

Do you see difference? OS's copy-paste does not change mtime, but changes ctime! But `cp` changes everything! 

So, if you asked me about rewriting tests using shell commands like `cp`, it does not seems possible. But if I don't understand you correctly, please give me explanation. Thanks!

phated:

Issue #193, opened 2016-07-21
Proposal to allow files to pass through `dest` without writing them to disk. 


doowb:
@jonschlinkert and I have been trying to figure out a way to pipe files through `dest` but not actually write the files to disk or create the directories.

The goal is to ensure the file properties are updated using the same logic from [prepare-write](https://github.com/gulpjs/vinyl-fs/blob/master/lib/prepare-write.js), then use the properties in another stream plugin. Our specific use case is to create a directory tree from the destination file paths.

We've come up with [this solution](https://github.com/doowb/vinyl-fs/tree/write-file), which uses a `file.writeFile` property to skip [creating the destination directory](https://github.com/doowb/vinyl-fs/blob/write-file/lib/prepare-write.js#L50-L53) and [return early in `write-contents`](https://github.com/doowb/vinyl-fs/blob/write-file/lib/dest/write-contents/index.js#L9-L12) when set to `false`.

Will you accept a pull request with this change, or suggest another way we could accomplish the same thing?

Thanks.

jonschlinkert:
fwiw we searched around quite a bit for other solutions/ideas before coming to this solution. 

I can't think of another way of being able to emulate the file tree - exactly as it would be if it were processed by `gulp.dest()`, but without actually writing the files. Every other way we came up with had some kind of drawback.

phated:
3 ideas come to mind.

* An option for dest called write to mirror read on src. 
* A separate method that just prepares the file. 
* (my favorite currently) Split the prepare logic into a module that exposes a stream and use that internally in vinyl-fs.

The reason I like the last one is that other vinyl adapters could use it to emulate our behavior.

jonschlinkert:
> The reason I like the last one is that other vinyl adapters could use it to emulate our behavior.

I love that idea. It might resolve/allow other things we've wanted to do.

doowb:
> Split the prepare logic into a module that exposes a stream and use that internally in vinyl-fs.

I think this is my favorite also. I was thinking `vinyl-prepare` or `vinyl-prewrite` for the name.

Also, I think that the `mkdirp` call should be moved to `write-contents`. I'll set the `dirMode` on the `file` so it can be used like this:

```js
var writeFolder = path.dirname(file.path);
fo.mkdirp(writeFolder, file.dirMode, onMkdirp);
```

If this sounds good, I'll put a repo and PR together.

contra:
[...]

PR #230, opened 2017-04-26
Refactor with vinyl-prepare

phated:
This is the WIP branch using vinyl-prepare.  I might also try to land #139 (including sourcemap support) in this changeset (separate commits).


erikkemperman:
Thanks for untangling this @phated! I'll try to keep things more cleanly separated in future.

About #139, on the src side, I believe every option except `sourcemaps` now supports function values, passing the Files as argument. Well, and `passthrough` isn't called per file, obviously.

I was wondering if it makes sense to also rename our `read` option so it no longer collides with through2?

phated:
@erikkemperman no problem.  I would typically just squash your stuff with the github UI but I wanted to document the extra features.

I need to review all the options stuff (I want to try to centralize it in some manner) before I close that issue.  I think you landed a rename (to `readFile`) in one of your commits.  However, I'm going to be removing the ability to pass options to `through2` because it was a hacky solution to us not using streams properly (now fixed).

erikkemperman:
I looked at centralizing them when I was looking at this, figuring prepare would be the place for it. But I got stuck, mostly because you preferred glob (well, file operations) stay at vinyl-fs. Perhaps a separate module used by vinyl-fs and vinyl-prepare?

Yeah, I renamed the option to readFile internally because we now need it later than we did before (per file instead of once up front) and as such  it would have clashed with through2.

phated:
@erikkemperman I've added the sourcemap(write) stuff into my first commit (since I'm no longer putting it in vinyl-prepare).  Can you take a look and give me feedback?  I'm going to be applying similar logic into your commit for src

erikkemperman:
Sure, I'll take a look. One potential downside, I guess, to not having this in prepare is that the sourcemaps will be excluded in the "dry run" scenario that got all this started...

phated:
@erikkemperman that is one of my goals.  @doowb didn't need sourcemaps in his request but I was trying to over-engineer (as I do).

erikkemperman:
@phated Fair enough, it's just that I would have expected a dry run to include them (especially when they're to be written separately).

Reading through the original [issue](https://github.com/gulpjs/vinyl-fs/issues/193), it also mentioned moving the `mkdirp` call into write-contents. Not sure why it would matter, but doing that would get rid of the save-file.js stage, which as it stands does nothing else.

And actually, didn't you consider making mkdirp a stream, at one point?

phated:
The `mkdirp` feels messy in `write-contents/` so I made it a separate stream step.  I think it's probably better to have a `mkdirp-stream` utility but I don't know the best way to do it.  I really need to find the proper abstraction levels before I split things up (and take on the maintenance responsibility).  The original issue was about removing all `fs.*` calls from the preparation logic so putting sourcemaps into vinyl-prepare breaks that contract.

erikkemperman:
Right. It'd be slightly messy in write-contents, that's true.

I don't know enough about the original use case to judge, but my understanding is they wanted to be able to reconstruct the directory structure as it would be created by `dest`. But perhaps it's alright if that does not include the (possibly distinct, I imagine) location of sourcemaps.

So far this looks good, but I'd need more time than I have right now to try this in an actual project.

phated:
@erikkemperman that's actually the opposite of what they wanted. haha.  Thanks for taking a look; I'm working on the `src` stuff now.

doowb:
Our main goal is to be able to use `vinyl-prepare` to calculate the correct destination for a file without doing any of the file system parts. This would include not creating the directory structure. We want to be able to setup the file's path properties before we do something like render templates so those properties can be used inside the templates for linking between pages.

I'll take a closer look at this when I get a chance and provide some feedback (if necessary). It might be a couple of days so don't wait on me when you're ready to merge :).

phated:
@erikkemperman I'm sticking a bunch of stuff in this PR that is related to the prepare stream changes but I'm not sure how I'm going rebase/merge them into master.

package.json
```diff
@@ -41,6 +41,7 @@
     "strip-bom-stream": "^1.0.0",
     "through2": "^2.0.0",
     "through2-filter": "^2.0.0",
+    "to-through": "^1.0.1",
```
terinjokes: indention

lib/src/index.js
```diff
@@ -1,21 +1,15 @@
 'use strict';
 
-var pumpify = require('pumpify');
-var through2 = require('through2');
 var gs = require('glob-stream');
-var duplexify = require('duplexify');
-var merge = require('merge-stream');
+var pumpify = require('pumpify');
+var prepare = require('vinyl-prepare');
```
terinjokes: is `pumpify` and `vinyl-prepare` even being used?

erikkemperman:
@phated The opposite? Maybe I should have said "predict" rather than "reconstruct" but I wasn't _that_ far off, Shirley :-)

erikkemperman:
I think the gulp-sourcemaps dependency can now go away, right?

I'll link this into an actual project or two tonight (GMT+2) and let you know how that goes!

phated:
@erikkemperman I think there are a bunch of deps I need to remove with these changes.  I'll double check.

phated:
As soon as these tests finish running, I'm going to call this PR done.  There are still things to do but I don't want to keep lumping them in here.

PR #231, opened 2017-04-27
Fix #210 Windows symbolic links requiring admin

### LINKED ISSUE #210: Symlinks on Modern Windows with latest node/npm.

phated:
This is just a reference PR for #211 because I'm not sure if the original author has time to finish this up.

TODO:
* [x] Remove new `link` method
* [x] Remove the added `src` tests because they don't do anything
* [x] Change author of the commit to give original author credit
* [x] Documentation
* [x] Don't allow the `relative` option with junctions
* [x] Need test for `forceJunctionDir` (and rename `useJunctions`) - Support function

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,8 @@ Returns a stream that accepts [vinyl] `File` objects, create a symbolic link (i.
 __Note: The file will be modified after being written to this stream.__
   - `cwd`, `base`, and `path` will be overwritten to match the folder.
 
+__Note: On Windows, directory links are created using Junctions by default. Use the `useJunctions` option to disable this behavior.__
+
 #### Options
 
 - Values passed to the options must be of the right type, otherwise they will be ignored.
@@ -260,7 +262,15 @@ Whether or not the symlink should be relative or absolute.
 
 Type: `Boolean`
 
-Default: `false`.
+Default: `false`
+
+##### `options.useJunctions`
+
+Whether or not a directory symlink should be created as a `junction`.
+
+Type: `Boolean`
+
+Default: `true` on Windows, `false` on all other platforms
 
 ##### other
 
--- a/lib/dest/write-contents/write-stream.js
+++ b/lib/dest/write-contents/write-stream.js
@@ -41,7 +41,7 @@ function writeStream(file, onWritten) {
     file.contents.removeListener('error', onComplete);
 
     // TODO: this is doing sync stuff & the callback seems unnecessary
-    // TODO: do we really want to replace the contents stream or should we use a clone
+    // TODO: Replace the contents stream or use a clone?
     readStream(file, complete);
 
     function complete() {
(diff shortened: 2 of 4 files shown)


afinnell:
Good call. I have been lacking the time. Is there an IRC, HipChat, Slack or other IM group the team is "normally" on?

phated:
@afinnell we have a private IRC channel but most of us idle in #gulpjs on freenode.  We'll likely be creating a team slack as the team grows.

Thanks for all the great information.  I might try to incorporate that into the docs.

phated:
Just noticed that `hardlink` doesn't need a `relative` option because it is creating a new file.

phated:
@afinnell if you have some time, can you review the new tests I added.  I think I used `.hardlink` properly but I'd like your review to verify.

phated:
Travis seems to be failing pretty hard with these tests.  Something with permissions that I need to look into (they were passing on my machine).

phated:
@afinnell going further down this rabbit hole - it seems that neither linux or windows allows hardlinks with directories as the first argument (but osx does for some reasons).  I'm not sure the best solution for this problem, any ideas?

phated:
Going to cc @erikkemperman too for the above ^

phated:
It seems like the best solution is to error if `file.isDirectory()`.  It's unfortunate that node isn't quite as cross-platform as it advertises.

phated:
Now that I'm reading the description that @afinnell wrote in the `hardlink` file, it seems like the goal of this method (linking directories on Windows) is unachievable.  I think we might just ship the `forceJunctionDir` option and call it done.

phated:
@contra @afinnell @erikkemperman if any of you have a chance to review this, I'd appreciate it because it's changed quite a lot over the course of today.

erikkemperman:
Unfortunately I won't be much help on Windows issues, having ditched it long ago.

The caveats [here](https://github.com/gulpjs/vinyl-fs/pull/231/files#diff-6c822274d928a75825fb98638c458970R30) strike me as pretty serious, though. Should they perhaps be in README as well?

phated:
I think the caveats are overly cautious. To put it into perspective: npm always uses them on Windows with no way to disable.

erikkemperman:
Alrighty then :-)

PR #235, opened 2017-05-03
Breaking: Remove ability to pass options to through2

phated:
Opening for visibility.  I've fixed the non-sinking problems which were the whole reason for this "feature".

@erikkemperman this is how I'm going to handle our `read` option problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,6 @@ Default: `false`
 ##### other
 
 Any glob-related options are documented in [glob-stream] and [node-glob].
-Any through2-related options are documented in [through2].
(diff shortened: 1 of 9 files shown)


erikkemperman:
Looks good to me, but then I have no idea why this was supported in the first place :-)

phated:
I can explain it but I'm afk today.

erikkemperman:
Never mind, no worries -- from where I'm standing it's just good to lose that ugly hack around `read`.

phated:
I amended my commit and merged locally.

PR #236, opened 2017-05-05
No need for realpath calls, just expose the right stat

erikkemperman:
Getting back to this stuff after a while, and as it now feels like things are nearing a milestone, I wanted to get a few things that I've played around with earlier off my chest.

Tried to keep commits tidy but figured I'd make a single PR for them, hope that's all right. If there's things in here you don't like, and cherry-picking turns out to be finicky, please do let me know and I'll amend accordingly.

### DIFF
--- a/lib/src/resolve-symlinks.js
+++ b/lib/src/resolve-symlinks.js
@@ -20,29 +20,13 @@ function resolveSymlinks(opt) {
         return callback(statErr);
       }
 
-      globFile.stat = stat;
-
       if (!stat.isSymbolicLink() || !resolveSymlinks) {
+        globFile.stat = stat;
         return callback(null, globFile);
       }
 
-      fs.realpath(globFile.path, onRealpath);
-    }
-
-    function onRealpath(realpathErr, filePath) {
-      if (realpathErr) {
-        return callback(realpathErr);
-      }
-
-      if (!globFile.originalSymlinkPath) {
-        // Store the original symlink path before the recursive call
-        // to later rewrite it back.
-        globFile.originalSymlinkPath = globFile.path;
-      }
-      globFile.path = filePath;
-
       // Recurse to get real file stat
-      resolveFile(globFile, enc, callback);
+      fs.stat(globFile.path, onStat);
     }
   }
 


lib/dest/index.js
```diff
@@ -14,8 +15,9 @@ function dest(outFolder, opt) {
 
   var saveStream = pumpify.obj(
     prepare.dest(outFolder, opt),
+    makeDirs(opt),
```
phated: `makeDirs` can't come before `sourcemap` because `sourcemap` might emit files with a different dirpath than the original files.

lib/dest/make-dirs.js
```diff
@@ -4,13 +4,12 @@ var through = require('through2');
 var valueOrFunction = require('value-or-function');
 
 var fo = require('../file-operations');
-var writeContents = require('./write-contents');
 
 var number = valueOrFunction.number;
 
-function saveFileStream(opt) {
+function makeDirs(opt) {
```
phated: This is fine, I can probably start to refactor a mkdirp-stream module out of here.

lib/src/index.js
```diff
@@ -23,6 +23,10 @@ function src(glob, opt) {
   opt.readFile = opt.read;
   opt.read = undefined;
 
+  // Leverage node-glob's statCache, conservatively
+  // TODO allow users to provide this option, at their own risk?
+  opt.statCache = Object.create(null);
```
phated: Not sure if I like this; I think it will cause more problems than it is worth.

lib/src/resolve-symlinks.js
```diff
@@ -1,52 +1,44 @@
...
 
-function resolveSymlinks(options) {
+function resolveSymlinks(opt) {
 
-  var resolveSymlinks = koalas(boolean(options.resolveSymlinks), true);
+  var resolveSymlinks = koalas(boolean(opt.resolveSymlinks), true);
 
   // A stat property is exposed on file objects as a (wanted) side effect
   function resolveFile(globFile, enc, callback) {
+    var stat = opt.statCache[globFile.path];
```
phated: I don't trust node-glob.  I've seen weird behavior with this.

lib/src/resolve-symlinks.js
```diff
@@ -1,52 +1,44 @@
...
-      if (!globFile.originalSymlinkPath) {
-        // Store the original symlink path before the recursive call
-        // to later rewrite it back.
-        globFile.originalSymlinkPath = globFile.path;
-      }
-      globFile.path = filePath;
-
       // Recurse to get real file stat
-      resolveFile(globFile, enc, callback);
+      fs.stat(globFile.path, onStat);
```
phated: Why was recursion removed and replaced with a `fs.stat` call (as opposed to `lstat`)

lib/src/resolve-symlinks.js
```diff
@@ -1,52 +1,44 @@
...
 
     function onStat(statErr, stat) {
       if (statErr) {
         return callback(statErr);
       }
 
-      globFile.stat = stat;
-
       if (!stat.isSymbolicLink() || !resolveSymlinks) {
+        opt.statCache[globFile.path] = stat;
```
phated: Again, don't trust node-glob

lib/dest/index.js
erikkemperman: Oh, good catch. Perhaps a test for that would be worthwhile. I figured this order was equivalent except it would fail fast when mkdir goes awry.

phated: There's likely many tests that need to be written around sourcemaps because I don't trust it's fully stable yet (as signified by the <1.0 version)

lib/src/index.js
erikkemperman: Agree about being wary of node-glob, I might get back to that later.

But the `statCache` implementation seems robust to me, in the sense that if we hit a match within it, those stats are reliable -- which is all I assume here.

It shouldn't increase risk of race conditions since we start with an empty object, and actually maybe decrease slightly because we keep stats consistent with what node-glob got while assembling matches.

But fair enough, let's forget about it for now.

phated: I'd feel better if we had more tests around all of this.

phated: Mind opening it as a separate PR and I'll flag it as "help wanted" and ask for help writing more tests.

lib/src/resolve-symlinks.js
erikkemperman: It took me a while to figure out what this was doing, but eventually I noticed that it was just a really roundabout way of obtaining the stats of the target of a symlink -- using `fs.realpath` followed by `fs.lstat`, followed by a hack to restore the file path to pre-realpath value.

But that is just what `fs.stat` does, and with that in place there is no need for realpath or recursion.

lib/src/index.js
erikkemperman: Sure, no problem.

lib/src/resolve-symlinks.js
phated: I could have sworn that `fs.stat` doesn't work in some case (so `lstat` was used).  Maybe @contra or @terinjokes remembers?

erikkemperman:
I've swapped makeDirs and sourcemaps back around, and removed the statCache commit for now.

lib/src/resolve-symlinks.js
erikkemperman: That would really reflect poorly on node fs... Anyway, note I still start with `lstat` (hidden in GH diff) and only proceed with a `stat` in case the first stat is a symbolic link and `resolveSymlinks` is true.

phated: > That would really reflect poorly on node fs

^ Remember that I had to create a custom mkdir method because node's `fs` doesn't create with proper permissions.

phated: @erikkemperman got some feedback:
> lstat works with links right
> if the file is a link it returns the stat for the actual file not the link

erikkemperman:
Ah, hadn't noticed you already merged something into master. Should I rebase this or leave it be for the moment?

phated:
@erikkemperman you can wait a little bit, I'm going to cherry-pick 9e7d8d5 into master when the tests run so we'll just have the `stat`/`lstat`/`realpath` commit to review.

lib/src/resolve-symlinks.js
erikkemperman: >  if the file is a link it returns the stat for the actual file not the link
That last bit describes `stat`, not `lstat`, I believe, and is consistent with node docs.

phated: Idk, I'm 99% sure there's weirdness you are missing.

erikkemperman:
All right, but 9e7d8d5 is the one you just merged, right? I think you might have meant https://github.com/gulpjs/vinyl-fs/pull/236/commits/ae8edcfb41cfbab265a02e8866b54ac4f735a9c7.

The remaining commit would need rebase, let me know if I can help.

phated:
@erikkemperman cherry-pick'd, please rebase so we can narrow the discussion.

lib/src/resolve-symlinks.js
erikkemperman: Possibly -- that's never something I could rule out :-) Let's look at it carefully. If there's any edge cases I missed (and the tests currently don't cover) then they should be plugged anyway, it seems to me...

phated: Oh!!!  Directories!!!  See https://github.com/gulpjs/vinyl-fs/blob/master/test/symlink.js#L244-L254

erikkemperman: Not sure what you mean, that test is still in there and passes with or without this change?

erikkemperman:
Rebased, now just the one commit. Thanks for merging the other stuff!

lib/src/resolve-symlinks.js
phated: Sorry, that's the example of one difference between `stat` and `lstat`:  `lstat` tells us the symlink is to a directory, while `stat` doesn't.  It's not a test for the changes you made (which would be `src` tests).

phated:
Approved the review since it's been rebased.  I'll keep poking at the 1 commit now.

lib/src/resolve-symlinks.js
erikkemperman: Yes, that's what I've been working with (and the node docs claim). So that would seem to confirm that my change of a second `stat` could replace the earlier detour via `realpath`+`lstat`, right?

erikkemperman:
Please do, thanks again for looking into these changes!

lib/src/resolve-symlinks.js
phated: Probably.  I'm also going to look into the recursion, because I think that's meant to traverse links to links to links to links, etc.

erikkemperman: I don't think the recursive structure was changed here, it's always at most one more trip -- after `stat` or `realpath` we can never end up with a stat that's `isSymbolicLink` a second time.

phated: Ah, perfect.  That was my biggest concern.  I think I might have come up with a test that breaks some assumptions, give me a minute.

erikkemperman: Well, technically I suppose I _did_ alter the recursive pattern, it is now recursing more tightly (just to the onStat callback) rather than the long way around (to the through2 transformation).

phated: I think I've thoroughly been schooled.  It seems that `fs.stat` does all resolution of symlinks in one step, as opposed to us traversing each symlink with `fs.lstat`.  @erikkemperman you never let me down!

phated:
Thanks a ton @erikkemperman! This change is going to speed things up considerably.

erikkemperman:
That was the plan, this should save a bunch of filesystem calls. Thanks for this rapid-fire interaction here, that really helps things along nicely.

It's still not ideal though, seeing largish numbers of redundant stats being done -- but I appreciate the worry about node-glob...

PR #237, opened 2017-05-05
Leverage node-glob stats cache

erikkemperman:
@phated As discussed elsewhere, a PR that might serve as discussion piece on the merits and perils of node-glob's cache(s).


erikkemperman:
Well that can't be good -- this passed earlier on my repo:
https://github.com/erikkemperman/vinyl-fs/commits/use-glob-cache
https://ci.appveyor.com/project/erikkemperman/vinyl-fs/build/42
Have I introduced some intermittent problem?

phated:
This happened a couple times on my runs today. I have no idea why it happens.

erikkemperman:
Similar error, node version, as this failure here? I've restarted my rebo build and it passes still:
https://ci.appveyor.com/project/erikkemperman/vinyl-fs/build/43

erikkemperman:
Oops, sorry, I was trying to re-start the Appveyor build of this PR but touched master. I should probably get some sleep.

phated:
Restarted the AppVeyor build on this.

lib/src/index.js
```diff
@@ -19,6 +19,10 @@ function src(glob, opt) {
     throw new Error('Invalid glob argument: ' + glob);
   }
 
+  // Leverage node-glob's statCache, conservatively
+  // TODO allow users to provide this option, at their own risk?
+  opt.statCache = Object.create(null);
```
phated: Should never mutate options that are passed in.

lib/src/resolve-symlinks.js
```diff
@@ -13,14 +13,22 @@ function resolveSymlinks(opt) {
 
   // A stat property is exposed on file objects as a (wanted) side effect
   function resolveFile(globFile, enc, callback) {
+    var stat = opt.statCache[globFile.path];
+    if (stat && (resolveSymlinks !== stat.isSymbolicLink())) {
```
phated: `resolveSymlinks !== stat.isSymbolicLink()` seems too clever.  I think it's more explicit `&&` between the two like below.

jonschlinkert:
> do either of you have experience with node-glob's statCache? It scares me.

If I remember correctly, an advantage of using the cache is that it can speed up subsequent lookups for the same files (mainly from "overlapping" glob patterns). Which doesn't seem like it would be problematic on the "first lookup", but it would need to be updated when files are moved, deleted or created - like in a watch scenario.  I don't recall how the stat cache works in node-glob, so my points might be totally irrelevant.

lib/src/index.js
erikkemperman: Ah, right. I figured it might be all right here as it's not a documented option, but yes it would linger on the options object of course. I propose to leave this as is for just now, pending the options refactor you've started.

lib/src/resolve-symlinks.js
erikkemperman: I think I see what you mean. I've been working with a branch of node-glob that would let me further reduce some fs calls, and with that branch it would be possible for the first stat we get from the cache to be symbolic. That's why this test currently catches both ways.

I'm not holding my breath on node-glob/minimatch PRs though, unfortunately the maintainer seems a little bit too busy to address issues. Some of which strike me as pretty serious (i.e. complete mismatch of path separator handling between glob and minimatch on windows) and I wonder if it might make sense to proceed with a fork of those libraries...

erikkemperman:
@jonschlinkert The way I'm using the cache here is reasonably safe I think, starting with an empty cache and then using any hits from it immediately, after glob-stream and before any other filesystem changes could occur -- well, within vinyl-fs anyway.

Having said that, I am still slightly anxious about the intermittently failing test on Appveyor, but if I understood @phated correctly he's seen that also without this patch?

erikkemperman:
This needs to be rebased (including the new options resolution), but I'm not sure how. My first thought is that `resolve-options` could somehow be made to operate across modules, and then `glob-stream` could make use of it too. I'll get back to this when I find more time...

phated:
@erikkemperman I've been running some benchmarks with your `use-cache` branch and it doesn't seem to improve the performance of src -> dest (I'm comparing master vs use-cache branch).

erikkemperman:
@phated I have a hunch the performance degradation is elsewhere, and might be sufficiently bad that it defeats measuring the impact of using this cache. Looking into that now...

phated:
@erikkemperman I agree that the main degradation is elsewhere, but this actually reduces the performance more (on top of that).  Hopefully we can figure it out.

erikkemperman:
Rebased, would depend on (something like) https://github.com/gulpjs/resolve-options/pull/3

(See [line 40 in package.json](https://github.com/gulpjs/vinyl-fs/pull/237/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R40))

phated:

phated:
@erikkemperman just noticed this was never re-opened after my rebase.  Still have interest in it?

erikkemperman:
@phated Ah, thanks for reminding me! Yeah, I looked into it briefly, trying to figure out exactly in which cases node-glob populates the stat cache.

Turns out it's only for patterns that don't have any wildcards in the last path component, e.g. `**/literal`, because only in that case will it do an `fs.stat` (actually an `lstat` first) to see if the file exists. Otherwise it'll do a `readdir` to get a list to match the wildcard patterns against ans will assume those files exist and not get any stats (unless you force it to with a flag).

I'd be happy to rebase, but would need something like https://github.com/gulpjs/resolve-options/pull/3 in place, if I remember correctly.

PR #238, opened 2017-05-07
Refactor symlink

erikkemperman:
Tried to dry up the `symlink` stuff a bit, as per some TODOs, but stopped short of touching the interaction with vinyl-prepare (which I think @phated mentioned having worked on).

Notes:
- The first commit is unrelated but addresses a quirk I ran into whilst preparing the later ones.
- The `overwrite` logic for symlinks here does an `unlink` if the option is true, I'm not sure what that would do to junctions on Windows.
- Perhaps the last commit is actually Breaking.

### DIFF
--- a/lib/dest/write-contents/index.js
+++ b/lib/dest/write-contents/index.js
@@ -7,6 +7,8 @@ var writeStream = require('./write-stream');
 var writeBuffer = require('./write-buffer');
 var writeSymbolicLink = require('./write-symbolic-link');
 
+var fo = require('../../file-operations');
+
 function writeContents(opt) {
 
   function writeFile(file, enc, callback) {
@@ -38,26 +40,13 @@ function writeContents(opt) {
     // This is invoked by the various writeXxx modules when they've finished
     // writing the contents.
     function onWritten(writeErr) {
-      if (isErrorFatal(writeErr)) {
+      if (fo.isFatalOverwriteError(writeErr, file.flag)) {
         return callback(writeErr);
       }
 
       callback(null, file);
     }
 
-    function isErrorFatal(err) {
-      if (!err) {
-        return false;
-      }
-
-      if (err.code === 'EEXIST' && file.flag === 'wx') {
-        // Handle scenario for file overwrite failures.
-        return false;
-      }
-
-      // Otherwise, this is a fatal error
-      return true;
-    }
   }
 
   return through.obj(writeFile);
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -2,19 +2,17 @@
 
 var fs = require('graceful-fs');
 
+var fo = require('../../file-operations');
+
 function writeSymbolicLink(file, onWritten) {
   // TODO handle symlinks properly
   fs.symlink(file.symlink, file.path, function(symlinkErr) {
-    if (isFatalError(symlinkErr)) {
+    if (fo.isFatalOverwriteError(symlinkErr)) {
       return onWritten(symlinkErr);
     }
 
     onWritten();
   });
-}
-
-function isFatalError(err) {
-  return (err && err.code !== 'EEXIST');
 }
 
 module.exports = writeSymbolicLink;
--- a/lib/symlink/index.js
+++ b/lib/symlink/index.js
@@ -11,6 +11,7 @@ var valueOrFunction = require('value-or-function');
 var koalas = require('koalas');
 var lead = require('lead');
 
+var fo = require('../file-operations');
 var makeDirs = require('../dest/make-dirs');
 
 var boolean = valueOrFunction.boolean;
@@ -60,31 +61,17 @@ function symlink(outFolder, opt) {
     }
 
     function onUnlink(unlinkErr) {
-      if (unlinkErr && unlinkErr.code !== 'ENOENT') {
-        return callback(unlinkErr);;
+      if (fo.isFatalUnlinkError(unlinkErr)) {
+        return callback(unlinkErr);
       }
       fs.symlink(srcPath, file.path, symType, onSymlink);
     }
 
     function onSymlink(symlinkErr) {
-      if (isErrorFatal(symlinkErr)) {
+      if (fo.isFatalOverwriteError(symlinkErr, file.flag)) {
         return callback(symlinkErr);
       }
       callback(null, file);
-    }
-
-    function isErrorFatal(err) {
-      if (!err) {
-        return false;
-      }
-
-      if (err.code === 'EEXIST' && file.flag === 'wx') {
-        // Handle scenario for file overwrite failures.
-        return false;
-      }
-
-      // Otherwise, this is a fatal error
-      return true;
     }
   }
 
--- a/lib/symlink/index.js
+++ b/lib/symlink/index.js
@@ -36,7 +36,7 @@ function symlink(outFolder, opt) {
     // * Most products CANNOT detect a directory is a Junction:
     //    This has the side effect of possibly having a whole directory
     //    deleted when a product is deleting the Junction directory.
-    //    For example, IntelliJ product lines will delete the entire
+    //    For example, JetBrains product lines will delete the entire
     //    contents of the TARGET directory because the product does not
     //    realize it's a symlink as the JVM and Node return false for isSymlink.
     var useJunctions = koalas(boolean(opt.useJunctions, file), (isWindows && isDirectory));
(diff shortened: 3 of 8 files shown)


lib/symlink/index.js
```diff
@@ -51,22 +51,24 @@ function symlink(outFolder, opt) {
...
 
-    // TODO: make DRY with .dest()
-    var dirMode = number(opt.dirMode, file);
-    var writeFolder = path.dirname(file.path);
-
-    fo.mkdirp(writeFolder, dirMode, onMkdirp);
+    // Because fs.symlink does not allow atomic overwrite option with flags, we
+    // delete and recreate if the link already exists and overwrite is true.
+    if (file.flag === 'w') {
+      // TODO What happens when we call unlink with windows junctions?
```
phated: If it deleted the fixtures on AppVeyor, the rest of the tests would fail, I think?

lib/symlink/index.js
```diff
@@ -51,22 +51,24 @@ function symlink(outFolder, opt) {
...
+      fs.unlink(file.path, onUnlink);
+    } else {
+      fs.symlink(srcPath, file.path, symType, onSymlink);
+    }
 
-    function onMkdirp(mkdirpErr) {
-      if (mkdirpErr) {
-        return callback(mkdirpErr);
+    function onUnlink(unlinkErr) {
+      if (unlinkErr && unlinkErr.code !== 'ENOENT') {
```
phated: Should this be abstracted like `isOverwriteError`?

lib/symlink/index.js
```diff
@@ -51,22 +51,24 @@ function symlink(outFolder, opt) {
...
+        return callback(unlinkErr);
       }
-
       fs.symlink(srcPath, file.path, symType, onSymlink);
     }
 
     function onSymlink(symlinkErr) {
-      if (isErrorFatal(symlinkErr)) {
+      if (fo.isOverwriteError(symlinkErr, file.flag)) {
```
phated: I noticed the `TODO` in `isFatalError` asked if we should check `file.flag`, did you find it appropriate to check that?

lib/symlink/index.js
erikkemperman: Just being cautious, I know nothing about junctions...

lib/symlink/index.js
erikkemperman: Thought about it, but couldn't see it being useful elsewhere just now.

lib/symlink/index.js
erikkemperman: Yeah, that's the possibly breaking change. Previously this ignored the flag. But I think this is what I'd expect the behavior to be (and note the tests were copied almost verbatim from the tests for `dest`)

lib/symlink/index.js
phated: Even though it's not being used another place, I think it would be clearer to someone reading the code in the future (and me, when I read it the first time).

lib/symlink/index.js
phated: Same, I'm not 100% on this either.

lib/symlink/index.js
phated: Alright, was this being tested then?  How could it if it was passing before?

lib/symlink/index.js
erikkemperman: @afinnell Do you happen to know, are have an opportunity to check, what happens when we call `fs.unlink` on a JUNCTION link on Windows? Hoping it just deletes the link, not the target...

lib/symlink/index.js
erikkemperman: All right, I'll amend accordingly.

lib/symlink/index.js
erikkemperman: Sorry, I guess that wasn't a very clear description on my part. So, the code for `symlinks` previously did not look at the `overwrite` option (or its derivative `file.flag` property) at all, and always assumed that an `EEXIST` error from `fs.symlink`could be ignored (as if `overwrite` were false).

The tests for `symlink` did not consider any overwrite scenarios (never attempted to symlink at an existing path) and so this never popped up.

So in this PR I've tried to account for `overwrite` with symlinks, and to make this consistent with how `dest` deals with it. My remark about having copied tests (almost) verbatim from `dest` was just to reinforce the last bit, that it now behaves (or it should) consistent with `dest`.

lib/dest/write-contents/index.js
```diff
@@ -38,26 +40,13 @@ function writeContents(opt) {
     // This is invoked by the various writeXxx modules when they've finished
     // writing the contents.
     function onWritten(writeErr) {
-      if (isErrorFatal(writeErr)) {
+      if (fo.isOverwriteError(writeErr, file.flag)) {
```
phated: I think this is the inverse of naming/usage.  I believe it's **not** an overwrite error if it returns `true`.  Maybe I'm wrong?

lib/dest/write-contents/write-symbolic-link.js
```diff
@@ -2,19 +2,17 @@
 
 var fs = require('graceful-fs');
 
+var fo = require('../../file-operations');
+
 function writeSymbolicLink(file, onWritten) {
   // TODO handle symlinks properly
   fs.symlink(file.symlink, file.path, function(symlinkErr) {
-    if (isFatalError(symlinkErr)) {
+    if (fo.isOverwriteError(symlinkErr)) {
```
phated: I think this is the inverse of naming/usage.  I believe it's **not** an overwrite error if it returns `true`.  Maybe I'm wrong?

lib/symlink/index.js
```diff
@@ -51,22 +51,24 @@ function symlink(outFolder, opt) {
...
+      fs.unlink(file.path, onUnlink);
+    } else {
+      fs.symlink(srcPath, file.path, symType, onSymlink);
+    }
 
-    function onMkdirp(mkdirpErr) {
-      if (mkdirpErr) {
-        return callback(mkdirpErr);
+    function onUnlink(unlinkErr) {
+      if (fo.isUnlinkError(unlinkErr)) {
```
phated: I think this is the inverse of naming/usage.  I believe it's **not** an unlink error if it returns `true`.  Maybe I'm wrong?

lib/symlink/index.js
```diff
@@ -51,22 +51,24 @@ function symlink(outFolder, opt) {
...
+        return callback(unlinkErr);
       }
-
       fs.symlink(srcPath, file.path, symType, onSymlink);
     }
 
     function onSymlink(symlinkErr) {
-      if (isErrorFatal(symlinkErr)) {
+      if (fo.isOverwriteError(symlinkErr, file.flag)) {
```
phated: I think this is the inverse of naming/usage.  I believe it's **not** an overwrite error if it returns `true`.  Maybe I'm wrong?

phated: This option will need to be added to the documentation (copy-paste from `.dest` options).

lib/dest/write-contents/index.js
erikkemperman: Yeah, now that you mention it, I agree. I took the "direction" of the original `isFatalError` functions but with the added specificity of the reasons (overwrite, unlink) it is kind of awkward. How about `emitOverwriteError` and `emitUnlinkError` or something like that?

phated: I would just invert.  Either name it `isNotOverwriteError` or do `!isOverwriteError` + invert true/false return values in the function.

lib/symlink/index.js
erikkemperman: Hmm, I thought I did that, but apparently not. Maybe didn't save. Will amend along with the error function names.

lib/dest/write-contents/index.js
erikkemperman: But that also seems slightly off, because the function returns true if it _is_ an overwrite error (`EEXIST`) but we want to ignore it (because `overwrite` was false)?

phated: That's why we want the negative version of it, right?  `isNotOverwriteError`?  I typically shy away from inverted function names but is seems clearer here.

erikkemperman:
@phated Sorry, didn't see your last remark in time, went with `isFatalOverwriteError` for now -- which is verbose but accurate, it seems to me. But let me know if I should change again, no problem at all!

phated:
I like that name

erikkemperman:
Cool. So the biggest open question, if I'm not missing anything, is whether or not we can rely on `unlink` not deleting the target for junctions on Windows. Unfortunately I don't have access to a Windows box to try this out.

And something I hardly looked at thus far is the logic in `lib/dest/write-contents/write-symbolic-link`...

lib/symlink/index.js
phated: It looks like Junction weirdness is only for directories and libuv errors if you try to unlink a directory symlink (https://github.com/libuv/libuv/blob/v1.x/src/win/fs.c#L707-L716)

phated: Should this option be mutually exclusive vs `useJunctions` like the `relative` option?

phated:
@erikkemperman This looks pretty good but I'd like for the `isFatal...` stuff to be a separate commit from the `overwrite` option addition so I can have a proper changelog.  Can you split that last commit?

erikkemperman:
@phated Sure, coming up.

lib/symlink/index.js
erikkemperman: I'm not too familiar with node source, but that block looks like a special case for dangling links, rather than junction links?

But again, I am utterly useless on Windows -- would really love for someone to just try this out (create junction same way as vinyl-fs, and then see what happens after `unlink`)...

phated:
Thanks @erikkemperman!  I'm going to merge this but make some nitpick changes to a few PRs (some rewording on the messages, etc)

erikkemperman:
Yup, thanks again for thorough review!

phated:
This has been merged into master (with a little rebase magic).

erikkemperman:
Rebased [revive-wrap-with-vinyl](https://github.com/erikkemperman/vinyl-fs/tree/revive-wrap-vinyl) and https://github.com/gulpjs/vinyl-fs/pull/237 for good measure.

PR #239, opened 2017-05-09
Option consolidation

phated:
@erikkemperman since you brought up option consolidation in that other thread, I wanted to get some thoughts put together (with this super hacky code) of a way we could consolidate options at the beginning of a pipeline and then make it available throughout `.src`/`.dest`/etc.

Lots of work still needs to be done (many TODOs inline) but I think this might be the best solution. Thoughts?


phated:
Using a map polyfill here turns a 15000ms test into a 51000ms test but there are ways around it (e.g. using `file.path` as the key).

erikkemperman:
@phated I am unable to time the polyfill case on Linux; if I run `TEST_PSEUDOMAP=true npm test` (or use an ancient node version) I get errors from the "integrations does not exhaust available file descriptors when streaming thousands of files" test... And some others as well, but I suspect those are actually residual errors from the first.

Any idea what I might be doing wrong?

erikkemperman:
I am having some more ideas but lack the time to play around with this any more just now.

For what it's worth, though, I've dropped in `es6-weak-map` instead of `pseudomap` and see no significant difference in `time npm test` between having `require('es6-weak-map')` versus `require('es6-weak-map/polyfill')` (on a node version that has native weakmap, obviously).

phated:
@erikkemperman even if you don't have the time to try your ideas, if you can send me the initial thoughts, it might help inform my work.

I really like the `es6-weak-map` polyfill but it does have some caveats (attaching a property to the Vinyl object) if WeakMap isn't natively supported.  This still might be a better solution than maintaining our own cache and adding a `uniqueId` to each vinyl object.

phated:
@erikkemperman I've updated to use an `options` API that wraps a WeakMap.  Thoughts?

erikkemperman:
Nodes >= 0.12 have native WeakMap, right? Perhaps it is alright to shim this, with the downside of extra vinyl property on 0.10, since it will go away eventually (and likely doesn't apply for most users today).

Quick thoughts, before I'm off to sleep:
- Maybe have separate weakmaps for src, dest, symlink invocations?
- Decouple define/declare options and their defaults from getting/coercing/using them, I think it is much cleaner to define stuff in a single place for each of the units that options are applied to, and this would also help correlating the docs to the source. 
- I expect lots of options, in common scenarios, to not actually vary from vinyl to vinyl, by virtue of not being functions (or functions with no args) or not being set (and getting constant defaults). It might save a lot of memory (*) to iterate over the options, before streaming, to identify such fixed options and keep those values in a separate (per-stream, rather than per-vinyl) map.
- So I was thinking three levels of map -- per invocation of export, per stream, per vinyl -- and possibly to generalize such a multiple-scope options utility, along with the coerce/coalesce magic, into a separate module for potential re-use in other (gulp) packages.

Does that make any sense at all? I think it might be fun to build such a thing.

(*) Not familiar with the map (object) implementation for node specifically, but in general re-allocating hash tables because of growth (which is repeated in case of poor initial estimates of typical capacity requirements) is an expensive operation, so it probably saves some time as well.

phated:
I cherry-picked the `revive-wrap-vinyl` changes and added the option resolution logic into `src` also.

phated:
Updated with `fs-mkdirp-stream` rebased in.

phated:
Trying to think of a good API for peeking on a pipeline to view each chunk as it enters and exits to clean this up.

phated:
@erikkemperman After talking to a few other people, I can't think of a good abstraction for this logic.  It's frustrating to not have an abstraction for you to work on your ideas but it might just have to be embedded in the project for now.

erikkemperman:
@phated I haven't found a lot of time for this recently... But just so I understand, you've included the vinyl-prepare logic back in the main module here -- so shouldn't the original use case of a `dest` dry-run still be addressed? If I remember correctly there was talk of a working version with a `write` option, analogous to `read` option to `src`?

phated:
I'm thinking about dropping that use case from the 3.0 goals to get this done.

phated:
This PR needs to be completely rethought in light of the discussion at https://github.com/gulpjs/vinyl-fs/commit/14d3fd5c5bb0d37d3db889f8aa459260c59baa04 😭

phated:
@erikkemperman completely rewritten with a new module (prototyping it as `hoptions`) powering the option resolution.  I still want to do a bit more cleanup on this but I think this solves the option resolution timing **and** my desire to define all the options in 1 place **without** having a terrible API.  Do you think you can implement your caching logic behind the `hoptions` API?

erikkemperman:
Ah, all right. It looks good at a first quick glance!

I had been messing about with further consolidating symlinks, but did so on top of the previous incarnation of options handling... Was hoping to clean that up and share today -- leaving for Italy tomorrow, yay! -- but won't  be able to rebase on top of this this evening, so that'll have to wait.

I probably won't be online much for about two weeks, so I may as well mention here something I ran into, although barely topical: I think `dest()` when given a symlink vinyl (e.g. as received from `src()` on a symlink with `resolveSymlinks == false`) should do the same junction voodoo on Windows as `symlink()`... Right?

But that requires `dest/write-contents/write-symbolic-link` to have stats for `file.symlink` to determine if the target is a directory, which we don't currently have in `dest()`. Not sure what the least ugly way around this is...

I almost wish I knew more about Windows, but if junctions only make sense for directories, can't we assume node will just correct `'junction'` type args to `fs.symlink` and use `'file'` instead, if the target isn't a directory?

BTW; I also think that `symlink()` should expose a `file.symlink` property on the vinyls it streams down to potential further stages, so that its symbolic vinyls are similar to those out of `src()`. Does that make sense?

phated:
@erikkemperman going to get this finished and merged today (I'm guessing you are still on vacation).  I think your previous comment actually has 2-3 different issues within it; can you create separate issues for each thing so we can revisit in the future?

PR #240, opened 2017-05-10
reintroduce prepareWrite back ontop of the options changes

phated:
@erikkemperman this is a concept of ditching `vinyl-prepare` given the options changes so things are consolidated once again.  Thoughts?


phated:
This is going to land with #239

PR #246, opened 2017-06-18
Dest symlinks

phated:
@erikkemperman does this address your thoughts from your last comment in #239?  `vfs.symlink` doesn't add the `.symlink` property yet (probably a separate thing to address).

A side thought: the `relative` option should probably renamed to `relativeSymlinks` for the `vfs.dest` method.

### DIFF
--- a/lib/dest/options.js
+++ b/lib/dest/options.js
@@ -1,5 +1,9 @@
 'use strict';
 
+var os = require('os');
+
+var isWindows = (os.platform() === 'win32');
+
 var config = {
   cwd: {
     type: 'string',
@@ -27,6 +31,18 @@ var config = {
   },
   sourcemaps: {
     type: ['string', 'boolean'],
+    default: false,
+  },
+  // Symlink options
+  useJunctions: {
+    type: 'boolean',
+    default: function(file) {
+      var isDirectory = file.isDirectory();
+      return (isWindows && isDirectory);
+    },
+  },
+  relative: {
+    type: 'boolean',
     default: false,
   },
 };
--- a/lib/dest/write-contents/index.js
+++ b/lib/dest/write-contents/index.js
@@ -12,6 +12,11 @@ var fo = require('../../file-operations');
 function writeContents(optResolver) {
 
   function writeFile(file, enc, callback) {
+    // Write it as a symlink
+    if (file.symlink) {
+      return writeSymbolicLink(file, optResolver, onWritten);
+    }
+
     // If directory then mkdirp it
     if (file.isDirectory()) {
       return writeDir(file, optResolver, onWritten);
@@ -20,11 +25,6 @@ function writeContents(optResolver) {
     // Stream it to disk yo
     if (file.isStream()) {
       return writeStream(file, optResolver, onWritten);
-    }
-
-    // Write it as a symlink
-    if (file.symlink) {
-      return writeSymbolicLink(file, optResolver, onWritten);
     }
 
     // Write it like normal
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -38,7 +38,6 @@ function writeSymbolicLink(file, optResolver, onWritten) {
     type: symType,
   };
 
-  console.log(srcPath, file.path);
   fo.symlink(srcPath, file.path, opts, onWritten);
 }
 
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -28,7 +28,7 @@ function writeSymbolicLink(file, optResolver, onWritten) {
 
   // This is done inside prepareWrite to use the adjusted file.base property
   if (isRelative && !useJunctions) {
-    srcPath = path.relative(file.base, file.symlink);
+    srcPath = path.relative(file.base, srcPath);
   }
 
   var flag = optResolver.resolve('flag', file);
@@ -38,6 +38,7 @@ function writeSymbolicLink(file, optResolver, onWritten) {
     type: symType,
   };
 
+  console.log(srcPath, file.path);
   fo.symlink(srcPath, file.path, opts, onWritten);
 }
 
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -1,18 +1,40 @@
 'use strict';
 
-var fs = require('graceful-fs');
-
 var fo = require('../../file-operations');
 
 function writeSymbolicLink(file, optResolver, onWritten) {
-  // TODO handle symlinks properly
-  fs.symlink(file.symlink, file.path, function(symlinkErr) {
-    if (fo.isFatalOverwriteError(symlinkErr)) {
-      return onWritten(symlinkErr);
-    }
+  var isDirectory = file.isDirectory();
 
-    onWritten();
-  });
+  // This option provides a way to create a Junction instead of a
+  // Directory symlink on Windows. This comes with the following caveats:
+  // * NTFS Junctions cannot be relative.
+  // * NTFS Junctions MUST be directories.
+  // * NTFS Junctions must be on the same file system.
+  // * Most products CANNOT detect a directory is a Junction:
+  //    This has the side effect of possibly having a whole directory
+  //    deleted when a product is deleting the Junction directory.
+  //    For example, JetBrains product lines will delete the entire
+  //    contents of the TARGET directory because the product does not
+  //    realize it's a symlink as the JVM and Node return false for isSymlink.
+  var useJunctions = optResolver.resolve('useJunctions', file);
+
+  var symDirType =  useJunctions ? 'junction' : 'dir';
+  var symType = isDirectory ? symDirType : 'file';
+  var isRelative = optResolver.resolve('relative', file);
+
+  // This is done inside prepareWrite to use the adjusted file.base property
+  if (isRelative && !useJunctions) {
+    srcPath = path.relative(file.base, srcPath);
+  }
+
+  var flag = optResolver.resolve('flag', file);
+
+  var opts = {
+    flag: flag,
+    type: symType,
+  };
+
+  fo.symlink(file.symlink, file.path, opts, onWritten);
 }
 
 module.exports = writeSymbolicLink;
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -1,8 +1,12 @@
 'use strict';
 
+var path = require('path');
+
 var fo = require('../../file-operations');
 
 function writeSymbolicLink(file, optResolver, onWritten) {
+  var srcPath = file.symlink;
+
   var isDirectory = file.isDirectory();
 
   // This option provides a way to create a Junction instead of a
@@ -24,7 +28,7 @@ function writeSymbolicLink(file, optResolver, onWritten) {
 
   // This is done inside prepareWrite to use the adjusted file.base property
   if (isRelative && !useJunctions) {
-    srcPath = path.relative(file.base, srcPath);
+    srcPath = path.relative(file.base, file.symlink);
   }
 
   var flag = optResolver.resolve('flag', file);
@@ -34,7 +38,7 @@ function writeSymbolicLink(file, optResolver, onWritten) {
     type: symType,
   };
 
-  fo.symlink(file.symlink, file.path, opts, onWritten);
+  fo.symlink(srcPath, file.path, opts, onWritten);
 }
 
 module.exports = writeSymbolicLink;
--- a/lib/file-operations.js
+++ b/lib/file-operations.js
@@ -243,6 +243,31 @@ function updateMetadata(fd, file, callback) {
   }
 }
 
+function symlink(srcPath, destPath, opts, callback) {
+  // Because fs.symlink does not allow atomic overwrite option with flags, we
+  // delete and recreate if the link already exists and overwrite is true.
+  if (opts.flag === 'w') {
+    // TODO What happens when we call unlink with windows junctions?
+    fs.unlink(destPath, onUnlink);
+  } else {
+    fs.symlink(srcPath, destPath, opts.type, onSymlink);
+  }
+
+  function onUnlink(unlinkErr) {
+    if (isFatalUnlinkError(unlinkErr)) {
+      return callback(unlinkErr);
+    }
+    fs.symlink(srcPath, destPath, opts.type, onSymlink);
+  }
+
+  function onSymlink(symlinkErr) {
+    if (isFatalOverwriteError(symlinkErr, opts.flag)) {
+      return callback(symlinkErr);
+    }
+    callback();
+  }
+}
+
 /*
   Custom writeFile implementation because we need access to the
   file descriptor after the write is complete.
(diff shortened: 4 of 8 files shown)


lib/dest/options.js
```diff
@@ -29,6 +33,18 @@ var config = {
     type: ['string', 'boolean'],
     default: false,
   },
+  // Symlink options
+  useJunctions: {
+    type: 'boolean',
+    default: function(file) {
+      var isDirectory = file.isDirectory();
+      return (isWindows && isDirectory);
```
erikkemperman: Maybe the default should just be a boolean value, set to `isWindows`? We have to AND the resolved option with `isDirectory(file)` per file anyway, because we probably shouldn't assume that the `useJunctions` option provided by the user always satisfies that restriction. Does that make sense?

lib/dest/write-contents/index.js
```diff
@@ -12,6 +12,11 @@ var fo = require('../../file-operations');
 function writeContents(optResolver) {
 
   function writeFile(file, enc, callback) {
+    // Write it as a symlink
+    if (file.symlink) {
```
erikkemperman: I wasn't sure here if the criterium should be existence of `file.symlink` and/or `file.isSymbolic()`. Ideally, if vinyl-fs is consistent with adding the `symlink` property, it should not matter I guess. But as things stand, the `file.symlink` test only catches the vinyls as produced by `src()` with falsy `resolveSymlinks` option (for this file), I believe.

lib/dest/write-contents/write-symbolic-link.js
```diff
@@ -1,18 +1,44 @@
...
-  fs.symlink(file.symlink, file.path, function(symlinkErr) {
-    if (fo.isFatalOverwriteError(symlinkErr)) {
-      return onWritten(symlinkErr);
-    }
-
-    onWritten();
-  });
+  var srcPath = file.symlink;
+
+  var isDirectory = file.isDirectory();
```
erikkemperman: It seems to me that the vinyls ending up here currently (excepting the mocked ones you added) would be piped in from `src()` with the `resolveSymlinks` set to false. But those vinyls will have the stats of the link, and  thus `file.isDirectory` here is not actually testing whether the _target_ is a directory but the _link_ (and of course `isDirectory()` will always return false for stats that are `isSymbolic()`).

Actually I think this is a problem in the implementation of `symlink()` as well: we can't be sure the vinyl has the stats of the target at the moment we need to know if the target is a directory. In fact we just do `new fs.Stats()` if the vinyl doesn't have stats, in `symlink/prepare`...

I suppose we should do an extra `fs.stat` on the target path, here and in `symlink` -- but only if we really need to know if it's a directory (i.e. on Windows and with the `useJunctions` option set to true)?

lib/dest/write-contents/write-symbolic-link.js
```diff
@@ -1,18 +1,44 @@
...
 function writeSymbolicLink(file, optResolver, onWritten) {
-  // TODO handle symlinks properly
-  fs.symlink(file.symlink, file.path, function(symlinkErr) {
-    if (fo.isFatalOverwriteError(symlinkErr)) {
-      return onWritten(symlinkErr);
-    }
-
-    onWritten();
-  });
+  var srcPath = file.symlink;
```
erikkemperman: Maybe this ought to be called `target`, as in the `fs.symlink` call? Personally I think `source` is a bit confusing in this context, I think of this as the _destination_ of the link :-)

lib/dest/write-contents/write-symbolic-link.js
```diff
@@ -1,18 +1,44 @@
...
+  }
+
+  var flag = optResolver.resolve('flag', file);
+
+  var opts = {
+    flag: flag,
+    type: symType,
+  };
+
+  fo.symlink(srcPath, file.path, opts, onWritten);
```
erikkemperman: Would it be possible to deduplicate the junction voodoo and do it within `fo.symlink`, now that we have that separate function which is called from both `dest()` and `symlink()`?

stevelacy:
@phated I cloned to the branch and ran the tests, several failed for me on Linux. All were to do with the filesystem returning errors. I can provide detailed information as required.

phated:
@stevelacy These all pass on travis, so I am wondering what is different on your system.  Not sure if it belongs in this PR and might be something we look at in person

lib/dest/options.js
phated: Yeah, this should be a separate PR though, as it's not as simple as changing the default

lib/dest/write-contents/index.js
phated: I'm nervous about using the `isSymbolic()` method within vinyl-fs methods because I tried to use it in some other code I was experimenting with and it wasn't acting as I thought it would.  I found out this was due to not having a valid `stat` before I used it but I think checking this property is a better way to handle it because `isSymbolic()` can be true without having this property (I think?)

lib/dest/write-contents/write-symbolic-link.js
phated: From a semantic standpoint, you are correct.  But naming it target inverts my expectation (as someone that doesn't link very often).  I'm fine with altering naming conventions (with proper documentation comments) but I don't want to hold up release of 3.0

lib/dest/write-contents/write-symbolic-link.js
phated: People can inject whatever they want in the middle of the pipeline, so they could do anything like what I've mocked (they can even use a different Vinyl source).  We can't couple `dest` to `src` in that way and are better with a more lax, yet slightly leaky abstraction (with the `.symlink` property and usage of `fs.Stats`).

That's a bit rambling; let me know if I should further clarify.

lib/dest/write-contents/write-symbolic-link.js
phated: I thought about that, but I'd rather ship this last feature to get 3.0 released and work on de-duplication in a future release.

lib/dest/options.js
erikkemperman: All right, that makes sense.

lib/dest/write-contents/index.js
erikkemperman: Yes, I believe `file.isSymbolic()` returns true if and only if `file.stats && file.stats.isSymbolicLink()`. So that doesn't work for stat-less files.

Relying on `symlink` here should be fine, I think, but like I mentioned earlier I think that property should be attached in more places than is currently being done.

I had that figured out already though so I could quickly make a PR for that once this one is merged?

phated: Yeah, that sounds great.  I should be merging this tomorrow.

lib/dest/write-contents/write-symbolic-link.js
erikkemperman: Fair enough, although I'm not sure what you mean by documenting it?

The variable has no external visibility and is only used briefly within the innermost scopes of `dest/write-contents/write-symbolic-link` and `symlink/index`?

lib/dest/write-contents/write-symbolic-link.js
erikkemperman: Sorry, my remark wasn't very clear :-)

I agree about the loose coupling, but my point here is that I think the `isDirectory` call here is _wrong_ in the common case of getting a symbolic vinyl as produced by `src()` with `resolveSymlinks` set to false!

Because it is looking at the wrong stats in that case, namely those of the link and not the target, and then `isDirectory` will return false even though the target might be a directory, and as a consequence the symlink type should've been`'junction'` but now won't be.

Does that help?

If you change the test for this, or add a new one, to work with an actually sourced symlink rather than one with a mock `isDirectory()`

erikkemperman: ... I think you will see what I mean. (Sorry, hit the button prematurely)

lib/dest/write-contents/write-symbolic-link.js
erikkemperman: All right, yeah it shouldn't be a breaking change.

erikkemperman:
@stevelacy Were your errors resolved now? I ask because, after nuking `node_modules` and doing a fresh `yarn install` I suddenly also get errors on previously working code. And both Travis and Appveyor fail now, even on master:

- https://ci.appveyor.com/project/erikkemperman/vinyl-fs/build/80/job/9n2xqwey8pfwoe0t
- https://travis-ci.org/erikkemperman/vinyl-fs/jobs/245354289

Will take a quick look to see if I can figure out what is going on.

stevelacy:
@erikkemperman @phated yes, those are exactly my errors. I could take a look at them once I get some free time

phated:
Seeing these on OSX too when deps updated.  Investigating

phated:
This looks like a problem with `readable-stream` because it seems to be the only dep that updated

phated:
@stevelacy @erikkemperman this is definitely a problem introduced in readable-stream as backported from node 8.  I'm working with them to get it fixed.

erikkemperman:
@phated Thanks for the update.

FYI I'm trying to get an extra test done this evening if I find some time, that might clarify the problem I believe still remains with `isDirectory()` in this PR.

phated:
@erikkemperman I'm pretty clear on what you are saying but you are thinking of it from an integration point-of-view.  I will post an example of why we can't make those assumptions (after I deal with this nodejs/streams issue).

erikkemperman:
@phated All right, I'll hang back for a bit :-)

lib/dest/write-contents/write-symbolic-link.js
phated: (Downloading the suggested patch so sending this now)

@erikkemperman The following example shows something a user might do that would require us to handle this case:

```js
vfs.src('./my-giant-image-dir')
  .pipe(through(function toSymlink(file, _, cb) {
    // Since a directory was sourced, file.isDirectory() will be true and we'll want to create a junction on Windows
    file.symlink = file.path;
  })
  .pipe(vfs.dest('./my-output-location'))
```

erikkemperman: @phated 
Yeah, I got that there may be cases where the vinyl will have non-symbolic stats as well as a `symlink` property, like your example<sup>*</sup>, and those look like they're being handled properly.

My thinking, poorly explained though it has been, was there may also be cases where the `isDirectory()` will return false even though the target of the symlink being created is actually a directory. For such cases I believe this PR will end up _not_ making a junction on Windows, which strikes me as incorrect?

E.g., just copying a link:
```js
vfs.src('./in/my-link-to-a-dir', { resolveSymlinks: false })
  .pipe(vfs.dest('./out'));
```

Does that make any sense? I guess I should've tried describing this in code earlier. 

*) Of course you were only making a point about `dest` here, but I guess such a case would usually just be piped into `vfs.symlink` without the need for an intermediate stage, right?

phated: 1. Yeah, my example is contrived but might be something people are using/would use because we didn't always have the `.symlink()` method.
2. You are totally correct; we need to think through a much better solution that what we are currently doing, but I think that requires my `better-stats` module that tracks more data than we currently get.  Until that time, it's probably best to expect users to handle those edge-cases themselves.
3. The other option would be to remove our `isDirectory` logic from the streams implementation and **only** use it as part of the `default`-er method so users can force junctions themselves.  Thoughts?

lib/dest/options.js
erikkemperman: Actually, isn't it as simple as changing the default? Seems to me that the directory case will be forced to symype `'file'` anyway, and the call to `file.isDirectory()` will have to be done in `write-contents/write-symbolic-link` in any event, so no need to do it twice?

Sorry if I'm missing something obvious, it's late here and been one of those days :-)

lib/dest/write-contents/write-symbolic-link.js
erikkemperman: The least invasive -- but pretty ugly -- way I see to handle both our examples would be to proceed with `isDirectory()` as you were, but if that returns false to do another `fs.stat`. E.g. something like

```js
  var isDirectory = file.isDirectory();
  if (!isDirectory && file.isSymbolic()) {
    isDirectory = fs.statSync(file.symlink).isDirectory();
  }
```
or somehow making that an async fs call.

Of course this is only ever an issue on Windows (the `type` arg to `fs.symlink` is ignored otherwise) and we'll want to avoid this complication entirely when not on Windows.

At one point I was thinking it might make sense to carry both regular and symbolic stats on symbolic vinyls. That would've made the recent issue with symlink resolution (the roundabout recursion which we managed to excise) a little less opaque, for example.

It'd be confusing to users to have two stats on vinyls though, so I guess it might only help within the bounds of vinyl-fs streams.

But if we don't keep both, notice that if we do indeed need an extra `fs.stat` here, it might be the _fourth_  stat/lstat shuffle for the same file in some scenarios: node-glob does an `lstat` and then a `stat`, we then do an `lstat` if resolveSymlinks is false in vfs.src() only to then follow it up here with another `stat` to implement junctions properly!

phated: Ugh, I really don't want to do that!  Maybe we ship this as a known bug with 3.0?

erikkemperman: Told you it was ugly :-P

The sync call was just to make it clear, it'd be async of course and only ever arise in a few unlikely cases on an already crappy OS.

I appreciate and share your impatience for release... But imho it would then be better to rip out junctions completely rather than having it half-baked -- especially for such a highly anticipated release?

By the way, I think I might have mentioned earlier: has anyone ever checked if node doesn't do this testing for directory targets for us, if it's in fact true that junctions don't work or make sense for anything other than directories? What happens if we pass 'junction' in an `fs.symlink` call if the target is a regular file?

stevelacy:
All tests passing on node 8.1.2 now

lib/dest/options.js
phated: We have to introduce `isDirectory` into the relative branch, I think.

lib/dest/write-contents/write-symbolic-link.js
phated: Just a code comment explaining what "target" means in the context.

lib/dest/write-contents/write-symbolic-link.js
phated: I'll give that a shot.

phated: Proved that Windows **does not** handle things properly if you force `'junction'` at https://ci.appveyor.com/project/gulpjs/vinyl-fs/build/487

phated: I don't think using another `fs.stat` is proper because the Vinyl object doesn't need to be produced from the filesystem, so that call would only be correct in some circumstances (yet still incorrect in others).

erikkemperman: True, but it would treat correctly it least the cases where it has the necessary information to, or so it seems to me.

It's too bad that node doesn't enforce this for us, but good to have that confirmed at least.

Having read up a little bit on junctions I'm wondering if the current approach, where it'd be the default even, may end up causing more problems than it solves?

As far as I can tell the problem was that symlinking dirs (but not files?) requires admin privileges.

But junctions are dangerous in that they may cause users to lose data (c.f. the caveats in the comments).

Then again that's easy for me to say because I don't use Windows.

phated: npm exclusively uses junctions on Windows, so I think the caveats were over-cautious.

phated:
@erikkemperman I believe I've created issues for the 4 outstanding items in this PR.  I think we should work on them in smaller chunks so I've merged this and we can work from there.

erikkemperman:
@phated Yup, thanks!

PR #252, opened 2017-06-23
Simplify `useJunctions` default

phated:
With this change, I think it was proper to change the `relative` logic to check against the `symType` and only reject the option if it is trying to generate a `junction`.

I also added more tests for the scenario I am solving with these changes.

/cc @erikkemperman


erikkemperman:
@phated Looks good to me!

PR #253, opened 2017-06-24
Update: attach symlink property (fixes #249)

### LINKED ISSUE #249: Attach symlink property to Vinyl objects when passed through `symlink` method
As per @erikkemperman's comments in #246, we should be attaching the `.symlink` property in other scenarios (other than just `src(..., { resolveSymlinks: false })`).  The most obvious of these is when the file passes through the `symlink` method and a symlink is created (I think this can be done in symlink's `prepare` stream to reduce some path history hackiness).

erikkemperman:
I ended up figuring the correct thing to do here is to have the `symlink` property be relative when created with the `relative` option set to true, and absolute otherwise. This mirrors the behaviour of `src` when reading links.

One slight quirk to notice is that we can't tell from just this property if the link is a junction: the trailing path separator it would have after `fs.readLink` is normalized off by the setter in vinyl. It's probably no big deal.

This is going to conflict with your other symlink PR, I expect, but it should be trivial to resolve.

### DIFF
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -5,8 +5,6 @@ var path = require('path');
 var fo = require('../../file-operations');
 
 function writeSymbolicLink(file, optResolver, onWritten) {
-  var srcPath = file.symlink;
-
   var isDirectory = file.isDirectory();
 
   // This option provides a way to create a Junction instead of a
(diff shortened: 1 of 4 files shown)


phated:
Cherry-picked as a39aeb891696295021dae0b750d7f0cb16ff1ca5

Thanks @erikkemperman

Issue #248, opened 2017-06-23
Review usage of `isDirectory` in `dest` symlink implementation


phated:
As per @erikkemperman's comments in #246, we need to review the way we are handling directories in the symlink path of `dest`.

erikkemperman:
And, depending on the outcome, possibly in `symlink` as well.

phated:
@erikkemperman I'm not sure if you are still on vacation but when do you think you'll have something for this?  We're getting *very* close to gulp 4 release - just a few dependency updates remaining and then this will become blocking.

I also wanted to ask if you think we could ship 3.0 without this and then if we had a working proposal before gulp 4, we could bump this to 4.0 and get it in gulp 4.

erikkemperman:
@phated I'm back -- just insanely busy at work, and on top of that my GF is moving in with me and has, um, ideas about decorating which somehow all involve me getting my powertools out :-)

I have some work in progress:
https://github.com/erikkemperman/vinyl/commits/symbolic-struct
https://github.com/erikkemperman/vinyl-fs/commits/symbolic-struct

Functionally I think it is an improvement, but boy is it ugly :-(

I've tried to add further integration tests for the other case I mentioned [here](https://github.com/gulpjs/vinyl-fs/pull/254#issuecomment-311197566) but those are failing.

You're welcome to take a look of course, perhaps you have some quick thoughts on improving it. Unfortunately I probably won't find a lot of time for this any time soon, probably not any at all for the next couple of days...

phated:
Thanks! As long as we have a starting point. (funny enough, I'm moving in with my gf right now too)

erikkemperman:
Forgot to answer your second question... This, or something like this, would be breaking in that the stats on the vinyl would now be symbolic after `symlink` or after `dest` given a symbolic vinyl.

IMHO the current behavior in that regard is incorrect (not just on Windows with junctions) and so it seems to me that ought to be addressed before 3.0 if at all feasible?

Having said that, I'm not too happy with the approach in the WIP I linked above... Perhaps we can come up with a way to remember both link and target stats within vinyl-fs (needed for handling junctions properly) but not expose the latter externally on symbolic vinyls?

erikkemperman:
By the way, congrats! Wish you two all the best!

phated:
Completed in #254

Issue #245, opened 2017-06-14
Symlink does not consider if a file has been modified


twifty:
This is one of those edge cases nobody likes to think about.

A few things I've noticed with symlinks:
1. If the file.basename is renamed before being piped into `symlink`, it will generate broken links.
2. If the file.contents is modified, a symlink is still generated (an error would be preferred).

Admittedly, I'm still using v2.4.4, but I don't see anything in master to suggest this has been fixed.

In my use case, I'm wrapping gulp to play nicely with a PHP framework. I've had to override `dest` to auto rename files and to log written filenames. One of the problems I'm trying to overcome is that if file contents have not been modified, they must be linked rather than copied. As it stands I have no way to detect this without also overriding `src` and possibly hashing the contents.

A great solution would be to add a `revision` property to the file object. This would start out at 0 and be increased any time the content is modified. This will also help solve problem (2) above.

In the mean time, can anybody suggest an ideal workaround for detecting changed contents?

erikkemperman:
Currently `symlink` just looks at the basename of the incoming vinyl file (the target of the symlink) and then creates a link with the same name in the destination directory. So if you change the basename (to something that doesn't exist) you will indeed get a broken symlink.

I guess it's reasonable to expect this to work, i.e. that the basename change only applies to the source but not the target of the link, but not sure how that could be implemented. I suppose we could examine the vinyl's `history` but that would be problematic with several renames (how far back in history should we go?).

We've looked at detecting changes to the contents, specifically to reflect accurate modification times in the vinyl's `stat` property:
https://github.com/gulpjs/vinyl/issues/72

But at the time I did not see a clean way to do this. It's not enough to look at just overwrites of the `contents` property (with a setter) because that won't catch modification of an existing buffer. Wouldn't want to monkeypatch that, I thought. Honestly if you want to be sure, checksum / hashing doesn't sound too bad?

twifty:
In my own implementation I simply added a checksum to the vinyl object and use the first `history` entry as the target. I'm guessing the performance shouldn't be an issue since most of the time, when specifying sources with symlinking in mind, one would pass the `buffer: false` option. But the file rename _is_ a problem. In a real world case, file names would have to be made url safe, which cannot be done with the current implementation of `symlink`.

I don't see how detecting the target would be a problem? wouldn't it always be the first history entry, resolved if that itself is a symlink? Any other filename in the chain cannot be trusted to point to a physical file/inode.

[...]

PR #254, opened 2017-06-26
Symlink/Junction improvements

phated:
As a talking point for #248, I wanted to create a couple of "integration" tests that show why we should keep the `file.isDirectory()` call inside `.dest` and `.symlink` methods.

@erikkemperman what do you think?

TODO: remove the `.only` call on the integration suite

### DIFF
--- a/lib/dest/options.js
+++ b/lib/dest/options.js
@@ -1,9 +1,5 @@
 'use strict';
 
-var os = require('os');
-
-var isWindows = (os.platform() === 'win32');
-
 var config = {
   cwd: {
     type: 'string',
@@ -34,13 +30,14 @@ var config = {
     default: false,
   },
   // Symlink options
-  useJunctions: {
-    type: 'boolean',
-    default: isWindows,
-  },
   relativeSymlinks: {
     type: 'boolean',
     default: false,
+  },
+  // This option is ignored on non-Windows platforms
+  useJunctions: {
+    type: 'boolean',
+    default: true,
   },
 };
 
--- a/lib/dest/prepare.js
+++ b/lib/dest/prepare.js
@@ -11,22 +11,23 @@ function prepareWrite(folderResolver, optResolver) {
   }
 
   function normalize(file, enc, cb) {
-    var mode = optResolver.resolve('mode', file);
-    var cwd = path.resolve(optResolver.resolve('cwd', file));
-
     var outFolderPath = folderResolver.resolve('outFolder', file);
     if (!outFolderPath) {
       return cb(new Error('Invalid output folder'));
     }
+    var cwd = path.resolve(optResolver.resolve('cwd', file));
     var basePath = path.resolve(cwd, outFolderPath);
     var writePath = path.resolve(basePath, file.relative);
 
     // Wire up new properties
-    file.stat = (file.stat || new fs.Stats());
-    file.stat.mode = mode;
     file.cwd = cwd;
     file.base = basePath;
     file.path = writePath;
+    if (!file.isSymbolic()) {
+      var mode = optResolver.resolve('mode', file);
+      file.stat = (file.stat || new fs.Stats());
+      file.stat.mode = mode;
+    }
 
     cb(null, file);
   }
--- a/lib/dest/write-contents/index.js
+++ b/lib/dest/write-contents/index.js
@@ -13,7 +13,7 @@ function writeContents(optResolver) {
 
   function writeFile(file, enc, callback) {
     // Write it as a symlink
-    if (file.symlink) {
+    if (file.isSymbolic()) {
       return writeSymbolicLink(file, optResolver, onWritten);
     }
 
--- a/lib/dest/write-contents/write-symbolic-link.js
+++ b/lib/dest/write-contents/write-symbolic-link.js
@@ -1,42 +1,74 @@
 'use strict';
 
+var os = require('os');
 var path = require('path');
 
 var fo = require('../../file-operations');
 
+var isWindows = (os.platform() === 'win32');
+
 function writeSymbolicLink(file, optResolver, onWritten) {
-  var isDirectory = file.isDirectory();
-
-  // This option provides a way to create a Junction instead of a
-  // Directory symlink on Windows. This comes with the following caveats:
-  // * NTFS Junctions cannot be relative.
-  // * NTFS Junctions MUST be directories.
-  // * NTFS Junctions must be on the same file system.
-  // * Most products CANNOT detect a directory is a Junction:
-  //    This has the side effect of possibly having a whole directory
-  //    deleted when a product is deleting the Junction directory.
-  //    For example, JetBrains product lines will delete the entire
-  //    contents of the TARGET directory because the product does not
-  //    realize it's a symlink as the JVM and Node return false for isSymlink.
-  var useJunctions = optResolver.resolve('useJunctions', file);
-
-  var symDirType =  useJunctions ? 'junction' : 'dir';
-  var symType = isDirectory ? symDirType : 'file';
-  var isRelative = optResolver.resolve('relativeSymlinks', file);
-
-  // This is done after prepare() to use the adjusted file.base property
-  if (isRelative && symType !== 'junction') {
-    file.symlink = path.relative(file.base, file.symlink);
+  if (!file.symlink) {
+    return onWritten(new Error('Missing symlink property on symbolic vinyl'));
   }
 
+  var isRelative = optResolver.resolve('relativeSymlinks', file);
   var flag = optResolver.resolve('flag', file);
 
-  var opts = {
-    flag: flag,
-    type: symType,
-  };
+  if (!isWindows) {
+    // On non-Windows, just use 'file'
+    return createLinkWithType('file');
+  }
 
-  fo.symlink(file.symlink, file.path, opts, onWritten);
+  fo.reflectStat(file.symlink, file, onReflect);
+
+  function onReflect(statErr) {
+    if (statErr && statErr.code !== 'ENOENT') {
+      return onWritten(statErr);
+    }
+
+    // This option provides a way to create a Junction instead of a
+    // Directory symlink on Windows. This comes with the following caveats:
+    // * NTFS Junctions cannot be relative.
+    // * NTFS Junctions MUST be directories.
+    // * NTFS Junctions must be on the same file system.
+    // * Most products CANNOT detect a directory is a Junction:
+    //    This has the side effect of possibly having a whole directory
+    //    deleted when a product is deleting the Junction directory.
+    //    For example, JetBrains product lines will delete the entire contents
+    //    of the TARGET directory because the product does not realize it's
+    //    a symlink as the JVM and Node return false for isSymlink.
+
+    // This function is Windows only, so we don't need to check again
+    var useJunctions = optResolver.resolve('useJunctions', file);
+
+    var dirType = useJunctions ? 'junction' : 'dir';
+    // Dangling links are always 'file'
+    var type = !statErr && file.isDirectory() ? dirType : 'file';
+
+    createLinkWithType(type);
+  }
+
+  function createLinkWithType(type) {
+    // This is done after prepare() to use the adjusted file.base property
+    if (isRelative && type !== 'junction') {
+      file.symlink = path.relative(file.base, file.symlink);
+    }
+
+    var opts = {
+      flag: flag,
+      type: type,
+    };
+    fo.symlink(file.symlink, file.path, opts, onSymlink);
+
+    function onSymlink(symlinkErr) {
+      if (symlinkErr) {
+        return onWritten(symlinkErr);
+      }
+
+      fo.reflectLinkStat(file.path, file, onWritten);
+    }
+  }
 }
 
 module.exports = writeSymbolicLink;
--- a/lib/file-operations.js
+++ b/lib/file-operations.js
@@ -155,6 +155,34 @@ function isOwner(fsStat) {
   return true;
 }
 
+function reflectStat(path, file, callback) {
+  // Set file.stat to the reflect current state on disk
+  fs.stat(path, onStat);
+
+  function onStat(statErr, stat) {
+    if (statErr) {
+      return callback(statErr);
+    }
+
+    file.stat = stat;
+    callback();
+  }
+}
+
+function reflectLinkStat(path, file, callback) {
+  // Set file.stat to the reflect current state on disk
+  fs.lstat(path, onLstat);
+
+  function onLstat(lstatErr, stat) {
+    if (lstatErr) {
+      return callback(lstatErr);
+    }
+
+    file.stat = stat;
+    callback();
+  }
+}
+
 function updateMetadata(fd, file, callback) {
 
   fs.fstat(fd, onStat);
@@ -413,6 +441,8 @@ module.exports = {
   getTimesDiff: getTimesDiff,
   getOwnerDiff: getOwnerDiff,
   isOwner: isOwner,
+  reflectStat: reflectStat,
+  reflectLinkStat: reflectLinkStat,
   updateMetadata: updateMetadata,
   symlink: symlink,
   writeFile: writeFile,
--- a/lib/symlink/link-file.js
+++ b/lib/symlink/link-file.js
@@ -1,50 +1,79 @@
 'use strict';
 
+var os = require('os');
 var path = require('path');
 
 var through = require('through2');
 
 var fo = require('../file-operations');
 
+var isWindows = (os.platform() === 'win32');
+
 function linkStream(optResolver) {
 
   function linkFile(file, enc, callback) {
-    var isDirectory = file.isDirectory();
-
-    // This option provides a way to create a Junction instead of a
-    // Directory symlink on Windows. This comes with the following caveats:
-    // * NTFS Junctions cannot be relative.
-    // * NTFS Junctions MUST be directories.
-    // * NTFS Junctions must be on the same file system.
-    // * Most products CANNOT detect a directory is a Junction:
-    //    This has the side effect of possibly having a whole directory
-    //    deleted when a product is deleting the Junction directory.
-    //    For example, JetBrains product lines will delete the entire
-    //    contents of the TARGET directory because the product does not
-    //    realize it's a symlink as the JVM and Node return false for isSymlink.
-    var useJunctions = optResolver.resolve('useJunctions', file);
-
-    var symDirType =  useJunctions ? 'junction' : 'dir';
-    var symType = isDirectory ? symDirType : 'file';
     var isRelative = optResolver.resolve('relativeSymlinks', file);
-
-    // This is done after prepare() to use the adjusted file.base property
-    if (isRelative && symType !== 'junction') {
-      file.symlink = path.relative(file.base, file.symlink);
-    }
-
     var flag = optResolver.resolve('flag', file);
 
-    var opts = {
-      flag: flag,
-      type: symType,
-    };
+    if (!isWindows) {
+      // On non-Windows, just use 'file'
+      return createLinkWithType('file');
+    }
 
-    fo.symlink(file.symlink, file.path, opts, onSymlink);
+    fo.reflectStat(file.symlink, file, onReflectTarget);
+
+    function onReflectTarget(statErr) {
+      if (statErr && statErr.code !== 'ENOENT') {
+        return onWritten(statErr);
+      }
+      // If target doesn't exist, the vinyl will still carry the target stats.
+      // Let's use those to determine which kind of dangling link to create.
+
+      // This option provides a way to create a Junction instead of a
+      // Directory symlink on Windows. This comes with the following caveats:
+      // * NTFS Junctions cannot be relative.
+      // * NTFS Junctions MUST be directories.
+      // * NTFS Junctions must be on the same file system.
+      // * Most products CANNOT detect a directory is a Junction:
+      //    This has the side effect of possibly having a whole directory
+      //    deleted when a product is deleting the Junction directory.
+      //    For example, JetBrains product lines will delete the entire contents
+      //    of the TARGET directory because the product does not realize it's
+      //    a symlink as the JVM and Node return false for isSymlink.
+
+      // This function is Windows only, so we don't need to check again
+      var useJunctions = optResolver.resolve('useJunctions', file);
+
+      var dirType = useJunctions ? 'junction' : 'dir';
+      var type = !statErr && file.isDirectory() ? dirType : 'file';
+
+      createLinkWithType(type);
+    }
+
+    function createLinkWithType(type) {
+      // This is done after prepare() to use the adjusted file.base property
+      if (isRelative && type !== 'junction') {
+        file.symlink = path.relative(file.base, file.symlink);
+      }
+
+      var opts = {
+        flag: flag,
+        type: type,
+      };
+      fo.symlink(file.symlink, file.path, opts, onSymlink);
+    }
 
     function onSymlink(symlinkErr) {
       if (symlinkErr) {
         return callback(symlinkErr);
+      }
+
+      fo.reflectLinkStat(file.path, file, onReflectLink);
+    }
+
+    function onReflectLink(reflectErr) {
+      if (reflectErr) {
+        return callback(reflectErr);
       }
 
       callback(null, file);
--- a/lib/symlink/options.js
+++ b/lib/symlink/options.js
@@ -1,13 +1,10 @@
 'use strict';
 
-var os = require('os');
-
-var isWindows = (os.platform() === 'win32');
-
 var config = {
+  // This option is ignored on non-Windows platforms
   useJunctions: {
     type: 'boolean',
-    default: isWindows,
+    default: true,
   },
   relativeSymlinks: {
     type: 'boolean',
@@ -16,12 +13,6 @@ var config = {
   cwd: {
     type: 'string',
     default: process.cwd,
-  },
-  mode: {
-    type: 'number',
-    default: function(file) {
-      return file.stat ? file.stat.mode : null;
-    },
   },
   dirMode: {
     type: 'number',
--- a/lib/symlink/prepare.js
+++ b/lib/symlink/prepare.js
@@ -1,7 +1,5 @@
 'use strict';
 
-// TODO: currently a copy-paste of prepareWrite but should be customized
-
 var path = require('path');
 
 var fs = require('graceful-fs');
@@ -13,7 +11,6 @@ function prepareSymlink(folderResolver, optResolver) {
   }
 
   function normalize(file, enc, cb) {
-    var mode = optResolver.resolve('mode', file);
     var cwd = path.resolve(optResolver.resolve('cwd', file));
 
     var outFolderPath = folderResolver.resolve('outFolder', file);
@@ -24,12 +21,16 @@ function prepareSymlink(folderResolver, optResolver) {
     var writePath = path.resolve(basePath, file.relative);
 
     // Wire up new properties
+    // Note: keep the target stats for now, we may need them in link-file
     file.stat = (file.stat || new fs.Stats());
-    file.stat.mode = mode;
     file.cwd = cwd;
     file.base = basePath;
+    // This is the path we are linking *TO*
     file.symlink = file.path;
     file.path = writePath;
+    // We have to set contents to null for a link
+    // Otherwise `isSymbolic()` returns false
+    file.contents = null;
 
     cb(null, file);
   }
--- a/test/dest-modes.js
+++ b/test/dest-modes.js
@@ -12,7 +12,7 @@ var statMode = require('./utils/stat-mode');
 var mockError = require('./utils/mock-error');
 var isWindows = require('./utils/is-windows');
 var applyUmask = require('./utils/apply-umask');
-var isDirectory = require('./utils/is-directory-mock');
+var always = require('./utils/always');
 var testConstants = require('./utils/test-constants');
 
 var from = miss.from;
@@ -135,7 +135,7 @@ describe('.dest() with custom modes', function() {
       path: inputDirpath,
       contents: null,
       stat: {
-        isDirectory: isDirectory,
+        isDirectory: always(true),
         mode: expectedMode,
       },
     });
@@ -164,7 +164,7 @@ describe('.dest() with custom modes', function() {
       path: inputDirpath,
       contents: null,
       stat: {
-        isDirectory: isDirectory,
+        isDirectory: always(true),
         mode: expectedMode,
       },
     });
@@ -251,7 +251,7 @@ describe('.dest() with custom modes', function() {
       base: inputBase,
       path: outputDirpath,
       stat: {
-        isDirectory: isDirectory,
+        isDirectory: always(true),
         mode: startMode,
       },
     });
@@ -259,7 +259,7 @@ describe('.dest() with custom modes', function() {
       base: inputBase,
       path: outputDirpath,
       stat: {
-        isDirectory: isDirectory,
+        isDirectory: always(true),
         mode: expectedMode,
       },
     });
--- a/test/dest-symlinks.js
+++ b/test/dest-symlinks.js
@@ -12,6 +12,7 @@ var vfs = require('../');
 var cleanup = require('./utils/cleanup');
 var isWindows = require('./utils/is-windows');
 var isDirectory = require('./utils/is-directory-mock');
+var isSymbolicLink = require('./utils/is-symbolic-link-mock');
 var testConstants = require('./utils/test-constants');
 
 var from = miss.from;
@@ -25,6 +26,11 @@ var outputPath = testConstants.outputPath;
 var inputDirpath = testConstants.inputDirpath;
 var outputDirpath = testConstants.outputDirpath;
 var contents = testConstants.contents;
+// For not-exists tests
+var neInputBase = testConstants.neInputBase;
+var neOutputBase = testConstants.neOutputBase;
+var neInputDirpath = testConstants.neInputDirpath;
+var neOutputDirpath = testConstants.neOutputDirpath;
 
 var clean = cleanup(outputBase);
 
@@ -33,11 +39,14 @@ describe('.dest() with symlinks', function() {
   beforeEach(clean);
   afterEach(clean);
 
-  it('creates symlinks when the `symlink` attribute is set on the file', function(done) {
+  it('creates symlinks when `file.isSymbolic()` is true', function(done) {
     var file = new File({
       base: inputBase,
       path: inputPath,
       contents: null,
+      stat: {
+        isSymbolicLink: isSymbolicLink,
+      },
     });
 
     // `src()` adds this side-effect with `resolveSymlinks` option set to false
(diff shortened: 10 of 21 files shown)


erikkemperman:
@phated

Interesting! If I were feeling contrary, couldn't I claim that this works just because the vinyl post-symlink is carrying the wrong stats (namely those of the target as opposed to the link)?

My point earlier, integration wise, was just this: shouldn't the output of `symlink` (and of `dest` having been fed a vinyl with `symlink` property) be a symbolic vinyl, with stats that return true for `isSymbolic()`?

 E.g., shouldn't
```js
vfs.src('./in/my-dir')
  .pipe(vfs.symlink('./tmp'))
  .pipe(vfs.dest('./out'));
```
be the same as doing the following, in series:
```js
vfs.src('./in/my-dir')
  .pipe(vfs.symlink('./tmp')),
vfs.src('./tmp/my-dir', {resolveSymlinks: false})
  .pipe(vfs.dest('./out'));
```
?

I might try playing around with something like your better-stats to see if I can't make it remember stats about both link and target. That might simplify this conundrum, it seems to me, whilst not suffering any  spurious calls (alternating between `fs.stat` and `fs.lstat` over the course of vfs internal pipe stages)...

phated:
@erikkemperman I had an idea before and forgot it but you just re-sparked it. What if `symlink` just becomes a bucket to store a target path and the stats you need later?  It's an undocumented property and we could mark it as a breaking change.  Does that solve the issues without taking on the complexity of `better-stats`?

erikkemperman:
That should work, I think!

phated:
@erikkemperman perfect!  Will you have time to put that together?  I'm a bit swamped this week and the little time I have I'm spending cleaning up the other modules for the 4.0 release.

erikkemperman:
@phated I'll give it a shot! I briefly tried something today but it got a bit messy, and I think an extra fs call can't be avoided (on windows) without tapping into mode-glob's cache.

Also got to wondering, what if an already symbolic vinyl is piped into `symlink`?

phated:
I think we just create a link to that link (a user might cause themselves to create a recursive link but that's always a risk with symlinks).

phated:
@erikkemperman where do we stand with this symlink stuff?  I agree that it's not semantically perfect but I think we can ship it (no one really uses these features, otherwise we'd have tons of bug reports because it barely works at all currently).  I just don't want us to get totally hung up on this as the only blocker.

erikkemperman:
@phated I tried your idea of making the `symlink` property hold both the path and stats but it got ugly for some reason. I'll look at it again asap.

phated:
@erikkemperman I'm adding more unit tests here before I take a look at the implementation.  I've just added the tests for emitting symbolic vinyls.  What other tests do you think we need here?

erikkemperman:
@phated Coincidentally, I was working on this earlier :-)

https://github.com/erikkemperman/vinyl-fs/tree/symlinks?files=1

I think this makes it functionally correct but it isn't especially pretty (an extra stat call in Windows).

In terms of tests I thought we could do with some more on how the vinyl looks post-`dest`.

Maybe we could merge ours efforts?

erikkemperman:
So I can't find where it was, but at some point I suggested we start with a simple, if ugly, approach that does the right things, and take it from there. That's what the commit I linked yesterday was attempting. I think the idea with making `symlink` property an object with both a path and stats might be unreliable -- although technically it seems possible as I've implemented it earlier, I felt it was quite brittle for some reason.

But looking back through some of these threads it seems I had repressed the complication you mentioned about users manipulating the vinyl, in particular its `isDirectory()` method, before it ends up in `symlink` or `dest`... And actually some of the unit tests have that mocked in, so it is slightly surprising that my branch passes both [Travis](https://travis-ci.org/erikkemperman/vinyl-fs/builds/281317210?utm_source=github_status&utm_medium=notification) and [Appveyor](https://ci.appveyor.com/project/erikkemperman/vinyl-fs/build/168).

Will need to look into this further :-(

erikkemperman:
Ok the tests pass without calling those mock `isDirectory()` functions because the paths just happen to point to actual directories... So that much at least makes sense.

erikkemperman:
Also, I wanted to add an integration test similar to yours but instead making the link with `symlink` rather than sourcing an existing symlink with `resolveSymlinks` set to false. However I will have to that with different paths because the one you used has different base names for link and target -- which `symlink` currently can't do. There's an [issue](https://github.com/gulpjs/vinyl-fs/issues/245) where that was discussed.

phated:
@erikkemperman I was working on some solutions based on your changes.  Something to note is that constructing a `new fs.Stat()` is generally undesirable because most properties become invalid when you construct without any arguments so we should always try to reuse what we have.

phated:
Also, I think I have a clean-ish solution that does the right thing in 99% of the cases and I'm fine not handling the 1% of edge cases; however, I want us to have a full test suite for everything we want to support **before** I work on the implementation.  Can you help with those remaining test cases?

phated:
Should `.dest` only write a symlink if `file.isSymbolic()`?  It currently checks for the `.symlink` property instead of `isSymbolic()`

phated:
@erikkemperman do you know if we can check flags to find out if something is a junction?  It looks like they set something on flags to *create* them

phated:
^ It looks like native code is required to `GetFileAttributes` on Windows where we'd be able to check for a "reparse point".  There's a native npm module that exposes this but I don't want to include it as a dependency just for Windows.

phated:
My solution is 4-fold:
1. `.src()` adds a "frozen" & "hidden" property (like `._symtype`) to each symlink Vinyl file using the logic from `.dest()`/`.symlink()` because it's actually loading a file into memory, the only time we can assume the original file exists. - I see this property eventually ending up in the `better-stats` module
2. `.dest()` should only enter into the `write-symbolic-link` path when `file.isSymbolic()` is true. This feels like the proper way it should be working and avoids us having to add extra `fs.Stat` logic in `dest/prepare` to ensure that a symlink is being emitted.
3. Both `.dest()` and `.symlink()` should attempt to read the `._symtype` property and only try to guess the symtype (using the current logic, including `file.isDirectory()`) if it doesn't exist on the Vinyl.  They would not read from the file system to guess this because we can't assume the file exists on disk.  If they need to guess the symtype, they would attach it as `._symtype` for future steps in the pipeline.
4. `.symlink()` should mask(?) the `stat.mode` with a symlink constant in `symlink/prepare` to ensure a symbolic Vinyl is emitted.

@erikkemperman please review the above and let me know if I'm missing anything.  I believe it handles your concerns and my concern about non-vinyl-fs-created Vinyl objects being streamed through the pipeline.  If this handles everything, what tests do we need to ensure the steps have full coverage?

erikkemperman:
@phated

About testing `.symlink` and/or `.isSymbolic()` in dest: we've discussed this earlier, [here](https://github.com/gulpjs/vinyl-fs/pull/246#discussion_r122681077). My takeaway from that was there might be cases where the vinyl might have `.symlink` but not have stats (and so `isSymbolic()` would return false). That's why I thought we should just check the former. But perhaps the safe bet would be `file.symlink || file.isSymbolic()`?

About detecting junctions: the tests currently look for a trailing slash out of `fs.readlink`. It seems to work but is kinda quirky.

Your 4-point strategy looks sound at a first casual glance, I might find time for a more thorough review, and perhaps suggestions on testing the whole thing, later today.

test/dest-symlinks.js
```diff
@@ -33,11 +34,14 @@ describe('.dest() with symlinks', function() {
...
   afterEach(clean);
 
-  it('creates symlinks when the `symlink` attribute is set on the file', function(done) {
+  it('creates symlinks when `file.isSymbolic()` is true', function(done) {
     var file = new File({
       base: inputBase,
       path: inputPath,
       contents: null,
+      stat: {
+        isSymbolic: isSymbolic,
```
erikkemperman: This seems wrong -- we'll want to mock either `file.isSymbolic()` or `file.stat.isSymbolicLink()` right?

https://nodejs.org/api/fs.html#fs_class_fs_stats
https://github.com/gulpjs/vinyl/blob/master/index.js#L84

test/dest-symlinks.js
```diff
@@ -59,6 +63,57 @@ describe('.dest() with symlinks', function() {
...
   });
 
+  it('does not create symlinks when `file.isSymbolic()` is false', function(done) {
+    var file = new File({
+      base: inputBase,
+      path: inputPath,
+      contents: null,
+      stat: {
+        isSymbolic: function() {
+          return false;
```
erikkemperman: Same as above.

test/dest-symlinks.js
```diff
@@ -59,6 +63,57 @@ describe('.dest() with symlinks', function() {
...
+    ], done);
+  });
+
+  it('errors if missing a `.symlink` property', function(done) {
+    var file = new File({
+      base: inputBase,
+      path: inputPath,
+      contents: null,
+      stat: {
+        isSymbolic: isSymbolic,
```
erikkemperman: Same as above.

test/dest-symlinks.js
```diff
@@ -59,6 +63,57 @@ describe('.dest() with symlinks', function() {
...
+      },
+    });
+
+    function assert(err) {
+      expect(err).toExist();
+      // TODO: Should we assert anything else about this err?
```
erikkemperman: FWIW, elsewhere we're testing the messages of (our custom) errors.

test/integration.js
```diff
@@ -49,4 +57,52 @@ describe('integrations', function() {
...
+      expect(files[0].symlink).toEqual(inputDirpath);
+    }
+
+    pipe([
+      vfs.src(inputDirpath),
+      vfs.symlink(symlinkDirpath),
+      vfs.dest(outputDirpath),
+      concat(assert),
+    ], done);
+  });
```
erikkemperman: I was thinking it might make sense to additionally test the case of directly sourcing a symlinked directory and copying it:

```js
it('(*nix) sources a symlink and copies it', function(done) {
    if (isWindows) {
      this.skip();
      return;
    }

    fs.mkdirSync(base);
    fs.mkdirSync(symlinkDirpath);
    fs.symlinkSync(inputDirpath, outputSymlink);

    function assert(files) {
      var destResult = fs.readlinkSync(outputDirpathSymlink);

      expect(destResult).toEqual(inputDirpath);
      expect(files[0].symlink).toEqual(inputDirpath);
    }

    pipe([
      vfs.src(outputSymlink, { resolveSymlinks: false }),
      vfs.dest(outputDirpath),
      concat(assert),
    ], done);
  });

  it('(windows) sources a junction and copies it', function(done) {
    if (!isWindows) {
      this.skip();
      return;
    }

    fs.mkdirSync(base);
    fs.mkdirSync(symlinkDirpath);
    fs.symlinkSync(inputDirpath, outputSymlink, 'junction');

    function assert(files) {
      // Junctions add an ending separator
      var expected = inputDirpath + path.sep;
      var destResult = fs.readlinkSync(outputDirpathSymlink);

      expect(destResult).toEqual(expected);
      expect(files[0].symlink).toEqual(inputDirpath);
    }

    pipe([
      vfs.src(outputSymlink, { resolveSymlinks: false }),
      vfs.dest(outputDirpath),
      concat(assert),
    ], done);
  });
```

erikkemperman:
Some addenda:
1. This new `_symtype` property would only need to be attached when the `resolveSymlinks` option is set to false, I guess?
2. I suppose the vinyls coming out of `symlink` and `dest` (if symbolic) should also have this new prop.
3. Should its (in)visibility and (im)mutability be tested?
4. Symlink shouldn't "mask" the mode but simply overwrite it to `parseInt('120777', 8)` (the lower 8 bits don't actually have meaning for links, but this is how they show up in `ls -la` after doing a regular `ln -s`).

erikkemperman:
Also, I'm wondering about re-using the `stats` in `symlink`. It is true that `new fs.Stats()` leaves things invalid but otherwise they'd be valid but _wrong_ (namely the target's metadata not the link's). Not sure which is worse?

erikkemperman:
Sorry for so many messages :-)

I've played around with the property you mentioned, but unless I'm missing something it's not getting us around the need for an extra `fs.stat` call on Windows, because we only ever end up where I think it would be attached (in `src/read-contents/read-symbolic-link`) if `resolveSymlinks` is false.

So at that point we'll only have the stats of the link, not the target, and other than (slightly awkwardly) detecting 'junction' by looking for a trailing slash out of `readlink` we can't tell the difference between 'file' and 'dir' without asking the filesystem.

If the above is accurate, then IMHO it's cleaner to do the stat call just before we need it, in `dest` rather than `src`, and we don't need the property I think.

For what it's worth, I've rebased by earlier effort on this branch, and the whole thing on latest master, and then updated to reflected the discussion here:
https://github.com/erikkemperman/vinyl-fs/tree/symlink-reb
(It passes all the tests.)

phated:
@erikkemperman I think our fundamental disagreement around this is that I am working under the ethos that `.dest()`/`.symlink()` should **not** be doing any reading of files that they haven't created.  Keeping the filesystem reads separate from the filesystem writes allows you to replace either end of the pipeline with a different vinyl adapter.  I think we can do those extra calls you mention in the `.src()` method and attach the results to the Vinyl object before we pass it along.  If we do it right, it'll only affect the performance on Windows in uncommon edge cases.

phated:
I think I'm missing a test that outlines my above philosophy.

phated:
> About testing .symlink and/or .isSymbolic() in dest: we've discussed this earlier, here. My takeaway from that was there might be cases where the vinyl might have .symlink but not have stats (and so isSymbolic() would return false). That's why I thought we should just check the former. But perhaps the safe bet would be file.symlink || file.isSymbolic()?
I think once all of this symlink stuff has been improved, it'll actually be correct to test for `isSymbolic()` but I'll just be keeping an eye on it for now.

phated:
> About detecting junctions: the tests currently look for a trailing slash out of fs.readlink. It seems to work but is kinda quirky.

I actually forgot about this.  I like it with a note about that caveat in the docs.

test/integration.js
phated: Added in 9c021e4

phated:
@erikkemperman just looking at your "symlink-reb" branch and I think that logic can be applied in `.src()` to reach my ideal solution.  I'm going to look at the tests again to embed my src = read/dest = write ethos.

test/dest-symlinks.js
```diff
@@ -270,6 +352,85 @@ describe('.dest() with symlinks', function() {
     ], done);
   });
 
+  // This could also be from a different Vinyl adapter
+  it('(*nix) receives a virtual symbolic directory and creates a symlink', function(done) {
```
phated: @erikkemperman does this help codify my ideology here?

test/dest-symlinks.js
```diff
@@ -270,6 +352,85 @@ describe('.dest() with symlinks', function() {
...
+
+    pipe([
+      from.obj([file]),
+      vfs.dest(neOutputBase),
+      concat(assert),
+    ], done);
+  });
+
+  // This could also be from a different Vinyl adapter
+  it('(windows) receives a virtual sybolic directory and creates a junction', function(done) {
```
phated: @erikkemperman same as above but for Windows.

phated: Ha, I guess this test represents my problem, as well as your's @erikkemperman - it didn't create a junction because it couldn't determine that the virtual directory was supposed to be a directory.  However, your solution doesn't solve this case because we can't look it up on the filesystem.  Ugh, I don't know how to resolve this.

phated:
@erikkemperman it's happy hour here, so I'm not sure if this is the best implementation **but** what do you think about https://github.com/phated/symlink-type/commit/e6a11b0d0d435de204a5f0d4e5e01c53fa955f79?? Maybe we can improve it to not error if it can't find the file and it could be used to achieve both our goals?

erikkemperman:
@phated I was thinking about your ethos that `dest` shouldn't do any reading. I believe I understand your point, and your use of the E word makes me think you feel quite strongly about it :-)

However, I do have some reservations... If the point is to be able to use different adapters for `src` and `dest`, then we're now going to be requiring alternative `src` adapters to also include this new `_symtype` property (only on Windows of course, it's  never needed otherwise)?

Also, at the point in `src` where this property would be attached, we won't know if we even need it later. So the fs call might be just wasteful. And if we do need it later, a bit of time will have passed since we checked the fs and thus the potential for changes in the meantime (either in the fs or in the vinyl attributes, e.g. a new path, making the property worse than useless but actual misinformation).

Not saying that doing the extra fs call (Windows) in `dest` is such a great idea, it obviously isn't, but I'm just not sure that doing it in `src` is actually improving things. And philosophically, this "read" is actually a part of making sure we do the write properly, and precisly in those cases where we need to be sure, so in that sense I guess I also feel it might be all right to do it in `dest`.

phated:
@erikkemperman I think I'm willing to just accept your https://github.com/erikkemperman/vinyl-fs/tree/symlink-reb patch.  It looks like it handles everything we want to and doesn't make it to hard to change in the future when we improve stats and stuff for Vinyl.  Do you think you can bring that together with this branch (I've just squashed everything and removed unnecessary stuff that I had added for `src`)

I can do some general cleanup if you want to submit it.

erikkemperman:
@phated Allright, I cherry-picked the stuff from my `symlink-reb` branch onto this one. Tests passing. I'm actually not too unhappy with where this ended up, only minor discomfort is that I'm still not sure what the correct (least bad) approach is for tacking stats on symbolic vinyls. Currently I just force the `mode` so that `isSymbolicLink` will return true -- but it will likely have the wrong uid/gid, size, atime/mtime.

lib/dest/options.js
```diff
@@ -36,7 +32,7 @@ var config = {
   // Symlink options
   useJunctions: {
     type: 'boolean',
-    default: isWindows,
+    default: true,
```
phated: Should add inline comment about this being ignored on non-Windows (as noted in the docs)

lib/dest/prepare.js
```diff
@@ -5,13 +5,15 @@ var path = require('path');
...
+var LINK_MODE = require('../constants').LINK_MODE;
+
 function prepareWrite(folderResolver, optResolver) {
   if (!folderResolver) {
     throw new Error('Invalid output folder');
   }
 
   function normalize(file, enc, cb) {
-    var mode = optResolver.resolve('mode', file);
+    var mode = !file.symlink ? optResolver.resolve('mode', file) : LINK_MODE;
```
phated: Should this check `file.isSymbolic()` like the other parts of the code?  I'm leaning towards yes but I guess we wouldn't need to change the mode then.

lib/dest/prepare.js
```diff
@@ -22,6 +24,8 @@ function prepareWrite(folderResolver, optResolver) {
     var writePath = path.resolve(basePath, file.relative);
 
     // Wire up new properties
+    // TODO symlinks: new stats will have invalid props, but target stats will
```
phated: Don't we read the `fs.stat` from the newly created file and update or something like that?

lib/dest/write-contents/write-symbolic-link.js
```diff
@@ -18,25 +25,41 @@ function writeSymbolicLink(file, optResolver, onWritten) {
...
-    file.symlink = path.relative(file.base, file.symlink);
+  if (isWindows) {
+    fs.stat(file.symlink, onStat);
+  } else {
+    onStat(null, file.stat);
   }
 
-  var flag = optResolver.resolve('flag', file);
+  function onStat(statErr, stat) {
+    if (statErr && statErr.code !== 'ENOENT') {
```
phated: Should we use a `isFatal*` method for this?

lib/dest/write-contents/write-symbolic-link.js
```diff
@@ -18,25 +25,41 @@ function writeSymbolicLink(file, optResolver, onWritten) {
...
+      return onWritten(statErr);
+    }
+    if (!stat) {
+      // If target doesn't exist, let's look at the vinyl to see which kind
+      // of dangling link we should to create...
+      stat = file.stat;
```
phated: This'll never work in this context.  We should probably just default to "file" (or undefined - which results in the same thing, I think) and document as such for dangling links.  Thoughts?

phated: Only suggesting this for `.dest()` because `.symlink()` behaves differently in this specific case.

lib/symlink/link-file.js
```diff
@@ -22,25 +24,41 @@ function linkStream(optResolver) {
...
 
-    var flag = optResolver.resolve('flag', file);
+    function onStat(statErr, stat) {
+      if (statErr && statErr.code !== 'ENOENT') {
+        return onWritten(statErr);
+      }
+      if (!stat) {
+        // If target doesn't exist, let's look at the vinyl to see which kind
+        // of dangling link we should to create...
+        stat = file.stat;
```
phated: I think this is proper here but not in the `.dest()` for symlinks.

lib/symlink/options.js
```diff
@@ -1,13 +1,9 @@
...
 
-var os = require('os');
-
-var isWindows = (os.platform() === 'win32');
-
 var config = {
   useJunctions: {
     type: 'boolean',
-    default: isWindows,
+    default: true,
```
phated: Inline comment like mentioned above

lib/dest/prepare.js
phated: Should the documentation for `.dest()` options.mode property be updated to note this?

phated: Does this even need to do this since we are **only** working with symbolic links for dest?

lib/symlink/prepare.js
```diff
@@ -24,8 +23,10 @@ function prepareSymlink(folderResolver, optResolver) {
     var writePath = path.resolve(basePath, file.relative);
 
     // Wire up new properties
+    // TODO new stats will have invalid props, but target stats will have the
+    // wrong ones...
     file.stat = (file.stat || new fs.Stats());
-    file.stat.mode = mode;
+    file.stat.mode = LINK_MODE;
```
phated: If we update the `fs.Stat` metadata after write in `.dest()` (don't remember), do we also do it for `symlink()`

test/dest-modes.js
```diff
@@ -135,7 +135,7 @@ describe('.dest() with custom modes', function() {
       path: inputDirpath,
       contents: null,
       stat: {
-        isDirectory: isDirectory,
+        isDirectory: isDirectory(true),
```
phated: Can't wait for `better-stats` so we don't need this pattern :frowning_face:

test/dest-symlinks.js
```diff
@@ -49,6 +59,8 @@ describe('.dest() with symlinks', function() {
       expect(files.length).toEqual(1);
       expect(file.symlink).toEqual(symlink);
       expect(files[0].symlink).toEqual(symlink);
+      expect(files[0].stat).toExist();
+      expect(files[0].stat.isSymbolicLink()).toBe(true);
```
phated: I don't like testing `isSymbolicLink` and would rather just test `isSymbolic` (the public interface)

test/dest-symlinks.js
```diff
@@ -74,6 +167,8 @@ describe('.dest() with symlinks', function() {
 
       expect(files.length).toEqual(1);
       expect(outputLink).toEqual(path.normalize('../fixtures/test.txt'));
+      expect(files[0].stat).toExist();
+      expect(files[0].stat.isSymbolicLink()).toBe(true);
```
phated: Same as above

test/dest-symlinks.js
```diff
@@ -281,7 +462,8 @@ describe('.dest() with symlinks', function() {
       path: inputDirpath,
       contents: null,
       stat: {
-        isDirectory: isDirectory,
+        isSymbolicLink: isSymbolicLink(true),
+        isDirectory: isDirectory(true),
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -245,7 +344,8 @@ describe('.dest() with symlinks', function() {
...
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -203,7 +301,8 @@ describe('.dest() with symlinks', function() {
...
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -167,7 +264,8 @@ describe('.dest() with symlinks', function() {
...
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -130,7 +226,8 @@ describe('.dest() with symlinks', function() {
...
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -94,7 +189,8 @@ describe('.dest() with symlinks', function() {
...
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -270,6 +370,87 @@ describe('.dest() with symlinks', function() {
...
+      return;
+    }
+
+    var file = new File({
+      base: neInputBase,
+      path: neInputDirpath,
+      contents: null,
+      stat: {
+        isSymbolicLink: isSymbolicLink(true),
+        isDirectory: isDirectory(true),
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -270,6 +370,87 @@ describe('.dest() with symlinks', function() {
...
```
phated: This can never happen

test/dest-symlinks.js
```diff
@@ -348,7 +534,8 @@ describe('.dest() with symlinks', function() {
       path: inputDirpath,
       contents: null,
       stat: {
-        isDirectory: isDirectory,
+        isSymbolicLink: isSymbolicLink(true),
+        isDirectory: isDirectory(true),
```
phated: This can never happen

test/symlink.js
```diff
@@ -111,6 +109,8 @@ describe('symlink stream', function() {
       expect(files[0].base).toEqual(outputBase, 'base should have changed');
       expect(files[0].path).toEqual(outputPath, 'path should have changed');
       expect(files[0].symlink).toEqual(outputLink, 'symlink should be set');
+      expect(files[0].stat).toExist('stat should be set');
+      expect(files[0].stat.isSymbolicLink()).toBe(true, 'stat should be symbolic');
```
phated: Same as the other refs to this

test/symlink.js
```diff
@@ -144,6 +144,8 @@ describe('symlink stream', function() {
...
```
phated: Same as the other refs to this

test/symlink.js
```diff
@@ -170,6 +172,8 @@ describe('symlink stream', function() {
...
```
phated: Same as the other refs to this

test/symlink.js
```diff
@@ -195,6 +199,8 @@ describe('symlink stream', function() {
...
```
phated: Same as the other refs to this

test/symlink.js
```diff
@@ -220,6 +226,8 @@ describe('symlink stream', function() {
...
```
phated: Same as the other refs to this

lib/dest/prepare.js
erikkemperman: Not sure I follow, sorry... `file.isSymbolic` only works later on _because_ we set mode here. If the vinyl doesn't have the `symlink` property (added by `src` or `symlink) the option resolver will fall back to using the mode already on the vinyl unless a custom one was explicitly given by the user.

I probably misunderstand your comment?

lib/dest/options.js
erikkemperman: Makes sense, yep.

lib/dest/prepare.js
erikkemperman: No, unlike the other write-xxx modules the symbolic ones (nor `symlink`) currently doesn't go through `fo.updateMetadata()`. Perhaps it should, partially, only to reflect the stats in disk.

lib/dest/write-contents/write-symbolic-link.js
erikkemperman: Maybe, sure.

lib/dest/write-contents/write-symbolic-link.js
erikkemperman: I only put this in to handle the case you suggested of users somehow manipulating the vinyl so it will return true for both `isSymbolic()` and `isDirectory()`, taking that as a cue for which kind of link to create in the absence of an actual target for it.

But simply using `'file'` in this case is fine with me.

lib/symlink/link-file.js
erikkemperman: Not sure I understand why it ought to be different?

lib/symlink/prepare.js
erikkemperman: We don't do it in either case (currently). Reason is probably we can't write stats for a symlink (but we could split the file-operations to make the reflecting the vinyl stats off disk part separately accessible)?

test/dest-symlinks.js
erikkemperman: Okay, sure.

test/dest-symlinks.js
erikkemperman: That's what I thought until you suggested we should handle users faking stuff like this :-) Did I misunderstand you (way) earlier?

lib/dest/prepare.js
phated: @erikkemperman theoretically, we are only receiving Vinyls with "stats" that match that of a Symbolic File (likely because someone used `resolveSymlinks: false` in `.src()` or forwarded by `.symlink()`) so I don't think we should be changing the mode here.  Maybe we just need to disallow the ability for a user to change the mode when `file.isSymbolic()` and note that in the documentation.  Thoughts?

lib/dest/prepare.js
phated: Yeah, I'm wondering if that helps a bunch of the problems we ran into with this implementation.

lib/dest/write-contents/write-symbolic-link.js
phated: @erikkemperman I've mostly given up on the use-case of synthetically-generated, incorrect Vinyl objects for this implementation.  I believe we can solve that desired use-case on the next iteration of this concept (when we have `better-stats`)

lib/symlink/link-file.js
phated: In the case someone does `vfs.src('my/cool/directory').pipe(vfs.symlink('my/link/location'))` - the `file.stat` property would be that of the thing (directory or file) we are creating a link for and we can rely on the stats.  The only case where that wouldn't work would be a symlink for a symlink (which would have to be dangling to not be handled by the code above this). 

This is opposite of `.dest()` where we will **only** end up in the symlink branch if the Vinyl is already symbolic.

lib/symlink/prepare.js
phated: I think that would be beneficial from the perspective of doing the same thing (or similar, in this case) for each operation.

test/dest-symlinks.js
phated: From above:
> I've mostly given up on the use-case of synthetically-generated, incorrect Vinyl objects for this implementation. I believe we can solve that desired use-case on the next iteration of this concept (when we have better-stats)

lib/symlink/prepare.js
```diff
@@ -24,12 +23,17 @@ function prepareSymlink(folderResolver, optResolver) {
...
     file.cwd = cwd;
     file.base = basePath;
     file.symlink = file.path;
     file.path = writePath;
+    // We have to set contents to null for a link
+    // Otherwise `isSymbolic()` returns false
+    file.contents = null;
```
phated: @erikkemperman this was an interesting find.  It makes sense, but I forgot we checked `isNull()` on the Vinyl

phated:
@erikkemperman I took care of a handful of the things.  I think we still need to tackle:
* Some sort of `updateMetadata` functionality
* The mode stuff for `.dest` when a Vinyl contains the `.symlink` property or `isSymbolic`
* A hefty amount of documentation updates to warn about all the caveats with this approach

Do you have time to look at any of that?

phated:
@contra @terinjokes @stevelacy can you guys review this solution?  I'd like to get some outside eyes on it.

test/integration.js
```diff
@@ -49,4 +60,133 @@ describe('integrations', function() {
...
+      this.skip();
+      return;
+    }
+
+    fs.mkdirSync(base);
+    fs.mkdirSync(symlinkDirpath);
+    fs.symlinkSync(inputDirpath, outputSymlink, 'dir');
+
+    function assert(files) {
+      // 'dir' symlinks add an ending separator
```
phated: @erikkemperman I guess our assumption about junctions adding the sep was incorrect.

erikkemperman: @phated Huh. Where did we get that assumption in the first place? Don't remember. Sorry I haven't gotten around to responding to the above...

lib/dest/prepare.js
stevelacy: I'll chime in, I agree a user shouldn't be able to change the mode.

nwhitmont:
Let me know if I can help move this along in any way.

lib/dest/prepare.js
erikkemperman: @phated Ah, now I see your point. Yes, if we let go of the scenarios of users manipulating vinyls, we can assume we only end up here directly from `src` with `resolveSymlinks: false` (meaning the vinyl will have the right stats, including mode) or `symlink` (which forces the mode to be symbolic). Then we can indeed simply test `file.isSymbolic()` and won't need to do anything here in that case.

lib/dest/prepare.js
erikkemperman: I'll try that.

erikkemperman:
@phated I made a couple of little changes, wasn't sure if I should pollute this thread (further) so I have them here for now:
https://github.com/erikkemperman/vinyl-fs/commits/is-directory-integrations
Feel free to cherry-pick if you like any of it...

But notice: there's a weird error, but only on Appveyor and only on Node 6. Superficially it looks like an overflow but need to check this out further... Unfortunately I'm going to have to call it a day.

erikkemperman:
@phated Pushed a few more changes pushed to the branch linked in the comment above. Mostly to use the new `reflectMetadata` more effectively.

I also figured out what is going wrong on Windows, see [this commit](https://github.com/erikkemperman/vinyl-fs/commit/e4088899a222a3f4b65c9fd43c21107710caeee4) which makes the build pass -- but I still need to track this down further to see what changed in the libuv source, I guess.

erikkemperman:
Reported the Windows quirk, FYI:
https://github.com/nodejs/node/issues/16496

phated:
Thanks for digging into that!

erikkemperman:
@phated I took a quick stab at documenting the caveats on dangling links on Windows, pushed it to my branch here:
https://github.com/erikkemperman/vinyl-fs/commits/is-directory-integrations

With something like that in place, I believe my branch now more or less covers all three bullets you listed [earlier](https://github.com/gulpjs/vinyl-fs/pull/254#issuecomment-338551482).

But having just written that caveats section, I wondered if instead of a (Windows only) `useJunction` option, perhaps it would be better to just hava a (Windows only) `symlinkType` option and just use that when we can't determine the right type (the target is not a regular file, but a directory or non-existent).

Thoughts?

phated:
@erikkemperman I'm getting back to this today.  I've cherry-picked some of your work but left out the 2nd round of `reflectMetadata` code.  I think we can use something similar but I'd like to clean it up a bit

lib/dest/prepare.js
```diff
@@ -11,22 +11,23 @@ function prepareWrite(folderResolver, optResolver) {
...
     var basePath = path.resolve(cwd, outFolderPath);
     var writePath = path.resolve(basePath, file.relative);
 
     // Wire up new properties
-    file.stat = (file.stat || new fs.Stats());
-    file.stat.mode = mode;
     file.cwd = cwd;
     file.base = basePath;
     file.path = writePath;
+    if (!file.isSymbolic()) {
```
phated: Need to document this in the options

lib/dest/write-contents/index.js
```diff
@@ -13,7 +13,7 @@ function writeContents(optResolver) {
 
   function writeFile(file, enc, callback) {
     // Write it as a symlink
-    if (file.symlink) {
+    if (file.isSymbolic()) {
```
phated: Not sure if the documents talk about `.symlink` but if so, it'll need to be updated

lib/file-operations.js
```diff
@@ -363,6 +391,7 @@ WriteStream.prototype.open = function() {
 // Use our `end` method since it is patched for flush
 WriteStream.prototype.destroySoon = WriteStream.prototype.end;
 // Use node's `fs.WriteStream` methods
+WriteStream.prototype._destroy = fs.WriteStream.prototype._destroy;
```
phated: @erikkemperman this is why your latest tests were failing for node 8 - they changed `.destroy` to `._destroy` on the prototype.

phated:
(Title changed to reflect what happened here)

@erikkemperman @stevelacy @contra could I get a thorough review of this?  It is the last thing blocking vinyl-fs 3.0.0 and I want to make sure it's done correct.

contra:
My only nitpick would be `junctions: true` instead of `useJunctions: true` as the option, but thats really just my personal taste.

erikkemperman:
@phated @contra @stevelacy I just had a bit of time to compare the diff of the current state of this PR against what I had in my branch, but superficially this looks good to me!

Could I just ask you guys to take a close look at the wording of the README changes I made, and which look like they made it in here verbatim? English isn't my first language :-) Specifically the caveats section about links on Windows. Thanks!

stevelacy:
Looks good, if I were to nitpick I'd add a link to the caveats section from the `useJunctions` bulletin point. `"Please refer to the [caveats section](<link>) below."`

phated:
@stevelacy great minds think alike - I'm going to do a doc review pass before I ship 3.0 and that was on my list.

@contra I think we talked about that in another issue/PR and decided to be more verbose for the Windows users.

@erikkemperman I actually fully reviewed the docs and found them great, which is why there was no comment from me 😛

phated:
Thanks everyone for the help - especially you @erikkemperman, most of this is your work 😄

erikkemperman:
@phated Excellent, very good to see this epic landed!

PR #257, opened 2017-07-05
Update: Support `encoding` option

erikkemperman:
Waiting for a bit of inspiration for the symlink/junction thing, I figured I might take a stab at https://github.com/gulpjs/vinyl-fs/issues/23.

It is a kind of "shallow" way to deal with encodings, limited to the filesystem terminals `src` and `dest` -- internally all buffers and streams are (still) UTF-8.

This has the advantage of (probably) not breaking any plugins, but incurs some needless (and in some scenarios perhaps even lossy) conversion when both source and destination files are in the same non-UTF8 encoding.


phated:
This is awesome! (though you might be insane if this was your time-biding activity 😉 ) I was approaching it from a completely different angle.  I'll do a full review (hopefully) soon.

phated:
@erikkemperman A question that just popped into my head: how does node's file writing methods handle their supported encodings?  If they handle it intrinsically, we can use [`Buffer.isEncoding(encoding)`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_isencoding_encoding) instead of your RegExp check, I think.

erikkemperman:
If I understand your question correctly, no... Because `iconv-lite` supports many more encodings than node does. This regex should match precisely those strings that iconv would canonize to `'utf8'`, here:

https://github.com/ashtuchkin/iconv-lite/blob/master/lib/index.js#L68

I thought about using native conversions for encodings that node supports, but decided against it because of added complexity (way more branches) and limited benefit (the fs methods, given an encoding, expect/return strings in stead of buffers, and I wanted to keep the vinyl API the same, so would need an extra conversion then) .

erikkemperman:
@phated I have made a couple of changes:
- Rebased (slight conflict in package.json).
- Got rid of that ugly regex, now just using iconv's method of canonizing encoding names.
- Added a bunch of BOM tests, and made the removeBOM option also apply to other BOM-aware encodings than UTF8.
- More tests all around.

lib/codecs.js
```diff
@@ -0,0 +1,23 @@
...
+
+var cache = {};
+
+function getCodec(encoding) {
+  var codec = cache[encoding];
+  if (!!codec || !encoding || cache.hasOwnProperty(encoding)) {
+    return codec;
+  }
+
+  try {
```
phated: Why do we want to swallow an unsupported codec?

lib/codecs.js
```diff
@@ -0,0 +1,23 @@
...
+    codec = iconv.getCodec(encoding);
+  } catch (err) {
+    // Unsupported codec
+  }
+
+  cache[encoding] = codec;
+  return codec;
+}
+
+getCodec('utf8');
```
phated: Can't we just do `var cache = { utf8: iconv.getCodec('utf8') }` above since we know it will be there?

lib/dest/write-contents/write-buffer.js
```diff
@@ -1,14 +1,30 @@
 'use strict';
 
+var iconv = require('iconv-lite');
+
 var fo = require('../../file-operations');
+var getCodec = require('../../codecs');
 
 function writeBuffer(file, optResolver, onWritten) {
+  var encoding = optResolver.resolve('encoding', file);
+  var codec = getCodec(encoding);
```
phated: Can `getCodec` (or a variation on it) be used as the "type" in the options instead of resolving a string then using this logic?

lib/dest/write-contents/write-buffer.js
```diff
@@ -1,14 +1,30 @@
...
+  if (encoding && !codec) {
+    return onWritten(new Error('Unsupported encoding: ' + encoding));
```
phated: Ah, this is how you are handling the "unsupported" case.  Interesting, hmmm.

lib/dest/write-contents/write-buffer.js
```diff
@@ -1,14 +1,30 @@
...
+
   var opt = {
     mode: file.stat.mode,
     flag: optResolver.resolve('flag', file),
   };
 
-  fo.writeFile(file.path, file.contents, opt, onWriteFile);
+  var contents = file.contents;
+
+  if (encoding && codec.enc !== 'utf8') {
```
phated: Codec is an object?  What other properties does it attach?

lib/dest/write-contents/write-buffer.js
```diff
@@ -1,14 +1,30 @@
...
+    contents = iconv.decode(contents, 'utf8', { stripBOM: false });
```
phated: If you decode to utf8 then encode as a different encoding, won't characters that utf8 doesn't support be changed/broken/something?

phated: Also, how do we know the original buffer was UTF8?  Node buffers can hold a few different encodings, I believe.

phated: An aside: does this logic make sense in `fo.writeFile`?  It might make it worth becoming a separate module (it already takes `opts`).

lib/dest/write-contents/write-stream.js
```diff
@@ -41,8 +60,10 @@ function writeStream(file, optResolver, onWritten) {
     file.contents.removeListener('error', onComplete);
 
     // TODO: this is doing sync stuff & the callback seems unnecessary
-    // TODO: Replace the contents stream or use a clone?
-    readStream(file, complete);
+    readStream(file, createResolver(readConfig, {
```
phated: This is painful. I understand why you are doing it, but damn.

lib/dest/write-contents/write-stream.js
```diff
@@ -1,9 +1,20 @@
...
+var createResolver = require('resolve-options');
+
 var fo = require('../../file-operations');
+var getCodec = require('../../codecs');
 var readStream = require('../../src/read-contents/read-stream');
+var readConfig = require('../../src/options');
 
 function writeStream(file, optResolver, onWritten) {
+  var encoding = optResolver.resolve('encoding', file);
+  var codec = getCodec(encoding);
```
phated: All the same questions as `write-buffer` apply to this file.

lib/src/read-contents/index.js
```diff
@@ -39,7 +39,10 @@ function readContents(optResolver) {
     // This is invoked by the various readXxx modules when they've finished
     // reading the contents.
     function onRead(readErr) {
-      callback(readErr, file);
+      if (readErr) {
```
phated: Nice consistency thing.  I'd like to see this split from this PR though.

lib/src/read-contents/read-buffer.js
```diff
@@ -1,23 +1,41 @@
...
-      file.contents = removeBomBuffer(data);
-    } else {
-      file.contents = data;
+    if (encoding) {
+      var removeBOM = codec.bomAware && optResolver.resolve('removeBOM', file);
+
+      if (codec.enc !== 'utf8') {
+        contents = iconv.decode(contents, encoding, { stripBOM: false });
+        removeBOM = removeBOM && contents[0] === '\ufeff';
+        contents = iconv.encode(contents, 'utf8');
```
phated: Inverse of my `write` question about translating to utf8.  Also, can node's `fs.readFile` read any encoding and store it in a buffer?  I thought they only supported a few encodings.  As you probably can tell, I don't understand this stuff very well.

lib/src/read-contents/read-buffer.js
```diff
@@ -1,23 +1,41 @@
...
+      }
+
+      if (removeBOM) {
+        contents = removeBomBuffer(contents);
```
phated: Do you not use iconv-lite's BOM removal to reduce branches and not decode/encode the common case of utf8?

lib/src/read-contents/read-stream.js
```diff
@@ -1,25 +1,36 @@
...
   file.contents = new lazystream.Readable(function() {
     var contents = fs.createReadStream(filePath);
 
-    if (removeBOM) {
-      return contents.pipe(removeBomStream());
+    if (encoding) {
+      var removeBOM = codec.bomAware && optResolver.resolve('removeBOM', file);
+
+      if (codec.enc !== 'utf8') {
+        contents = contents
```
phated: Same questions on this file as each other.

lib/codecs.js
erikkemperman: I figured we would cache the failed cases too -- that way we won't have to go into a try/catch repeatedly. The main thing is this function returns undefined when no codec is found, which is then used by the callers to emit an actual error.

Perhaps this new file needs a few tests as well, I might find some time today.

lib/codecs.js
erikkemperman: Yeah, that'd be a little bit less messy, agreed. Will amend.

lib/dest/write-contents/write-buffer.js
erikkemperman: Do you imagine folks providing their own implementation of iconv's codec API, as the value to the `encoding` option here? I guess that might work, with a few more changes, but seems a pretty far-fetched scenario... Much more likely they would just submit their codec to iconv, right?

Also, notice that we only use this codec object for two properties, namely `enc` and `bomAware`. The public API of iconv expects its arguments to `encode` and `decode` to be strings.

lib/dest/write-contents/write-buffer.js
erikkemperman: The only alternative to emitting the error would be to quietly ignore invalid `encoding` values and then proceed as if it were `'utf8'`. But that might cause some nasty bugs, and pretty hard to track down just from the end results (wrong bytes in the buffer).

lib/dest/write-contents/write-buffer.js
erikkemperman: Yes, an object. The exact details seems to depend on the `encoding` used to obtain it. For instance, not all of them actually have an `enc` property with the canonized name. I've verified, though, that the `'utf8'` one does (and I only use it to check that one case).

It's all slightly weird, the `getCodec` method is part of the public API but the return value isn't documented:
https://github.com/ashtuchkin/iconv-lite/blob/master/lib/index.js#L63

I should probably add a bunch of sanity checks (in `test/integration` I guess?) to verify that iconv behaves as expected, in case future versions of it violate our assumptions.

lib/dest/write-contents/write-buffer.js
erikkemperman: Yes, any characters that are not supported by the target encoding will be replaced with `'�'` (multi-byte encodings) or `'?'` (single-byte encodings).

https://github.com/ashtuchkin/iconv-lite/blob/master/lib/index.js#L15

That's just an inescapable consequence of different encodings including different characters, though. 

Having said that I believe that Unicode covers the vast majority of characters that were previously part of various older sets. For the vast majority of those that are not exactly replicated it has decent look-alikes / variants, and it should be the core business of something like iconv to leverage those instead.

For example, I am using an old Simplified Chinese set in the unit tests which predates Unicode, and that works out just fine (albeit for a short sentence).

Since UTF8 can reflect all of Unicode, it is certainly our best bet for internal/default encoding.

Of course Node buffers can contain other encodings, in fact it can contain raw binary stuff that's not encoded text at all. So if folks fill the buffer with non-UTF8 contents *and* then attempt to decode that to text, garbage will ensue. My contention is that that's not something vinyl-fs can, or should, protect against.

For what it's worth, it's still possible to read/write pure binary files with vinyl-fs. I took care to allow the `encoding` option to be `'false'` to do no transcoding at all (and also in that case it won't strip BOMs). I forgot to document that possibility, come to think of it, and I should also add separate tests for it.

lib/dest/write-contents/write-stream.js
erikkemperman: Yeah, ouch, indeed. I looked briefly into avoiding this whole mess -- i.e. for a shortcut to re-source a stream, rather than actually going through `src/read-contents/read-stream`...

But then I figured that even if I found something it probably shouldn't be in this PR. The current master branch does it this way, albeit without the added complexity of the option resolver. So I propose we look into this issue separately?

lib/src/read-contents/index.js
erikkemperman: All right, will do.

lib/src/read-contents/read-buffer.js
erikkemperman: I hope I've addressed the translation question adequately, and I think it applies here as well.

As for the `fs.readFile` thing, it's similar to `fs.writeFile`: Node has a horrible API, in my humble opinion, in its efforts to "help us" :-)

https://nodejs.org/dist/latest-v8.x/docs/api/fs.html#fs_fs_readfile_path_options_callback
https://nodejs.org/dist/latest-v8.x/docs/api/fs.html#fs_fs_writefile_file_data_options_callback

If we omit the `encoding` argument, these methods will yield/expect (readFile/writeFile respectively) a Buffer containing just the raw unadulterated bytes. However if we provide an encoding argument, then suddenly these methods yield/expect Strings, which it decodes/encodes (readFile/writeFile) for us.

Long story short, better to not pass `encoding` to the `fs` methods and rely on iconv to do this for us.

lib/src/read-contents/read-buffer.js
erikkemperman: Yeah, I was doing that in my first stab at this. But then I realized that we'd want to strip BOMs off of UTF16 encoded streams as well, and possibly other multi-byte encodings that support serialization with variable endianness.

So the best way to do that, I thought, was to transcode to UTF8 whilst explicitly _not_ stripping BOMs. At that point, all the various BOM sequences will be mapped to the familiar UTF8 BOM, which we were already equipped to deal with via `remove-bom-buffer` and `-stream`.

But we do take advantage of the decoding/encoding logic understanding about BOMs here: by peeking at the first character of the string in between decode/encode, I can "hint" whether there is a BOM to strip.

Does that make sense?

lib/src/read-contents/read-stream.js
erikkemperman: Let me know if there's anything I haven't addressed properly above!

lib/dest/write-contents/write-buffer.js
erikkemperman: Sorry, I misunderstood, I now realize you were saying `getCodec` could be used as the coercer function passed to value-or-function via resolve-options.

That makes a lot of sense, yeah. I'll try if I can get that to work, nice exercise also to see about error handling in resolve-options, which I wanted to do anyway.

erikkemperman:
@phated Pushed a couple of changes, which I think are pretty solid. I'm still looking into leveraging those codec objects a bit more effectively, including whether they could be woven into the option resolution.

By the way, I looked briefly at cloneable-readable, wondering if it might help resolve the situation that write-stream has to loop around to read-stream with option resolution and everything. I could not get that to work though, it appears both clones have to be sinked or neither will flow... Which is not what we want here, it seems to me -- and actually, isn't that a problem in vinyl as well?

erikkemperman:
@phated One more commit, I'll leave this alone for a while now, for a fighting chance at review :-)

Mostly I thought it might make sense to have all interaction with iconv in a single file, and test separately if it does what we want.

I may have gone slightly overboard with tests, actually.

I played around with your suggestion of using the `getCodec` function as option coercer, but either I am missing some obvious or I am running into some limitations (imho) of resolve-options/value-or-function... I have some ideas about that, but propose to leave it out of this PR.

erikkemperman:
@phated Rebased to resolve minor conflict.

phated:
@erikkemperman I've read through the code (but not the tests) and have some ideas floating around in my head.  However, I'd like to leave this until 3.1.0 since it is only adding a feature and not breaking anything (this is mostly because the 3.0.0 changelog is already going to be huge).  Do you agree?

P.S. I'm going to mark my previous review as dismissed but I don't think we've reached completion here.

lib/codecs.js
```diff
@@ -0,0 +1,105 @@
+'use strict';
+
+/* eslint-disable new-cap */
+
+var iconv = require('iconv-lite');
+var through = require('through2');
+
+var DEFAULT_ENCODING = require('./constants').DEFAULT_ENCODING;
+
+function Codec(codec, encoding) {
```
phated: I really like where this went with this abstraction but I think it can be improved upon.  I'm still formulating these ideas in my head.

erikkemperman: There's always room for improvement, curious what you're thinking!

erikkemperman:
@phated Yeah, it shouldn't be a breaking change. But on the other hand, if it's ready and stable why not include it -- the already big changelog wouldn't get that much bigger with this in, it seems to me.

I appreciate and agree it should not block anything, but maybe let's see where this PR is at when you feel 3.0 is otherwise mature?

phated:
@erikkemperman Sorry if I was unclear.  My statements were more about me not wanting to put time into my ideas until we wrap up the absolute blockers for 3.0 (symlinks, mainly).  If something comes to me randomly, I'll definitely work on this and if it lands before 3, that will be fine.

erikkemperman:
@phated All righty. By the way, I've been looking at the symlink (well, junction) stuff again but am now slightly doubtful we can safely re-use stats in `dest` -- I think it might be a decent idea to get something working that's at least functionally correct, even if it's ugly, and then we can always try to improve on that. Does that sound like a plan?

phated:
Sounds good!

erikkemperman:
@phated Rebased this for good measure. I would love to see something like this in 3.0, but you mentioned you had ideas for improvement and I'd totally understand if you prefer to keep this on the shelf for the time being.

phated:
While I think this is beneficial for the API, I want to spend my time focused on the large changeset we currently have and ensure that it works for everyone.  We don't have nearly enough eyes on the 3.0 changes.

phated:

erikkemperman:
Done, https://github.com/gulpjs/vinyl-fs/pull/287

Issue #208, opened 2016-09-01
Test failures on 32bit node - issues with timestamps


CurryKitten:
Running the test suite via npm test on 32bit node versions is failing on -

1) .dest() with custom times calls futimes when an mtime is provided on the vinyl stat:
2) .dest() with custom times calls futimes when provided mtime on the vinyl stat is valid but provided atime is invalid:
3) .dest() with custom times writes file atime and mtime using the vinyl stat:
4) updateMetadata updates times on fs and vinyl object if there is a diff:
5) updateMetadata updates the mode & times on fs and vinyl object if there is a diff:

They all fail in a similar way - which is a difference in the expected vs the actual timestamp, for example -

```
Uncaught Error: Expected 1472731813622 to equal 1472731813519
      + expected - actual

      -1472731813622
      +1472731813519
```

Running the 64bit version of node on the same machine passes the tests.  Is this expected to pass on 32bit ?

phated:
That's weird.  When you say 32bit node versions, what do you mean exactly?  Is that on a 32-bit OS?  What is the filesystem type? Etc.

CurryKitten:
The test machines are running 64 bit OS's, I'm running the 32-bit and 64-bit node binaries, i.e -

https://nodejs.org/dist/v4.5.0/node-v4.5.0-linux-x86.tar.xz
https://nodejs.org/dist/v4.5.0/node-v4.5.0-linux-x64.tar.xz

Filesystems are ext4.  As I say I can run both versions of node on the same machine/environment and the 32bit version throws up this odd error.

OpenGG:
Test enviroment:

- Debian 8
- ext4 file system

Tested Node.js versions:

- 0.10.48-x86
- 4.5.0-x86
- 8.1.2-x86

Results:

- The test cases @CurryKitten mentioned fail in v4.5.0-x86. The mtime passed to `fs.futimes()` and the one read from `fs.statsSync()` drift by a few hunderds of milliseconds, less than a second.
- But they pass in v0.10.48-x86 and v8.1.2-x86.

Conclusion:

Maybe an old bug in Node.js, and now fixed? Not sure yet, maybe I will do a test with major Node.js versions.

It is not likely a bug in vinyl-fs, I suggest a rewrite to the test cases, change

```js
console.error(futimesSpy.calls[0].arguments[2].getTime());
expect(futimesSpy.calls.length).toEqual(1);
[...]

PR #258, opened 2017-07-06
Fix utime-related test cases

OpenGG:
- Use `new Date()` instead of `fs.statSync(__filename).mtime` as `file.stat.mtime`
- Fix all cases mentioned in https://github.com/gulpjs/vinyl-fs/issues/208
- Make a note about precision lost in 32bit Node.js


phated:
This could probably have a test helper, but I think it's fine how it is.  I'm going to rebase the name of the first commit since I only use "Fix" when a bug is fixed in the library code and use "Update" or "Build" for test changes.

Thanks a ton for looking into this and sending a fix.

erikkemperman:
Oh crap, I started a review yesterday but apparently left it "Pending"... I was just going to nitpick that technically the problem, as I understand it, isn't casting (as the docs now state) but the loss of precision is due to unit conversion (millis to secs, dividing by 1000).

phated:
@erikkemperman Feel free to submit any changes. I'm otherwise archiving this.

PR #259, opened 2017-07-07
Read-contents/index, invoke cb with error or result, not both

erikkemperman:
Separate PR, as per @phated's comment on https://github.com/gulpjs/vinyl-fs/pull/257.

### DIFF
--- a/lib/src/read-contents/index.js
+++ b/lib/src/read-contents/index.js
@@ -39,7 +39,10 @@ function readContents(optResolver) {
     // This is invoked by the various readXxx modules when they've finished
     // reading the contents.
     function onRead(readErr) {
-      callback(readErr, file);
+      if (readErr) {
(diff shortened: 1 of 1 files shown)


phated:
Thanks @erikkemperman

phated:
@erikkemperman There seems to have been a race condition with node 0.10 on this change.  Not sure why but I re-ran and it passed.

erikkemperman:
@phated That's weird -- I don't see how this little change would do that... Do you have logs so I can take a peek? I've run the test suite a couple of times with 0.10.0 and 0.10.48 but haven't been able to reproduce it here (Linux).

phated:
Doesn't look like travis gives me access to the old logs of a restarted job.  It was something with a file descriptor being undefined when we expected it to be defined.  I'll keep an eye out in the future.

PR #260, opened 2017-07-13
Docs: reword 32-bit timestamp precision issue

erikkemperman:
As per https://github.com/gulpjs/vinyl-fs/pull/258#issuecomment-314924964


phated:
Thanks @erikkemperman!

PR #267, opened 2017-08-01
scaffold: Switch Travis.CI to use "Trusty" distro

BurtHarris:
Travis.CI builds are [warning](https://travis-ci.org/gulpjs/vinyl-fs/jobs/259003527):

> This job ran on our Precise environment, which is in the process of being decommissioned. Please read about the rollout of Trusty as default on the blog, and take note that you can explicitly stay on Precise by specifying dist: precise in your .travis.yml. 

This PR updates the **.travis.yml** configuration to move to their next-generation distro, "Trusty".   I'm no Linux expert, but It seems to work fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 sudo: required
 language: node_js
 node_js:


erikkemperman:
@BurtHarris Thanks! The fact that this PR passes Travis with the new distro is certainly good to know. 

However, it is my understanding that the recommended approach is to only explicitly specify a particular distro in the `.travis.yml` when you want to keep using an *older* distro than the default?

https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming

BurtHarris:
Strictly speaking this may not be necessary.   I think they will switch it over by default, since no distro is currently specified.

erikkemperman:
Yeah, I noticed it as well but figured it'd be safer to leave it alone (but keeping it in mind if I got errors I could not explain otherwise) because the risk is we'd effectively miss out on their next distro upgrade if we specify "trusty" by force. Does that make sense?

BurtHarris:
What would really make the most sense is omitting both distro and sudo.

erikkemperman:
@BurtHarris Almost missed that you added a bunch of stuff to your post after I'd replied already...

I think Travis are just saying that they will continue supporting the _combination_ of (explicit) `dist: precise` and `sudo: required` but no longer update the (precise) images. This will not affect us since we only have `sudo` but no `dist` (so we'll automatically be moved onto trusty infrastructure).

The not-owned fixtures are needed to test if vinyl-fs correctly handles files not owned by the process owner. To do this we need to `chown` these fixtures and that requires sudo obviously. This file is in the repository and therefore not in `.gitignore`. It is supposed to "hang around" after testing, but I haven't seen it show up in `git diff` after running the tests -- but I'm on Mac and Linux only, what's Windows saying has changed to this file?

BurtHarris:
I think `test/fixtures/not-owned/not-owned.txt` doesn't exist before running the tests, but does after (on Windows.)    I think that's because the test is skipped on Windows, but not before creating the file, but apparently the cleanup doesn't run if skipped.

erikkemperman:
https://github.com/gulpjs/vinyl-fs/blob/master/test/fixtures/not-owned/not-owned.txt

BurtHarris:
OK.   It's perhaps its a lineEnding issue.   when I run git diff after the test, I get 

```
warning: LF will be replaced by CRLF in test/fixtures/not-owned/not-owned.txt.
The file will have its original line endings in your working directory.
```

BurtHarris:
I'm ok with closing the PR without merging, your choice.

erikkemperman:
Cool, thanks again for submitting this though!

PR #270, opened 2017-08-02
Upgrade CI scripts and eslint

BurtHarris:
- Adds latest stable build of nodejs to test matrix
- Run eslint only on latest stable build of nodejs, not older builds
- Upgrade eslint to latest
- remove obsolete jscs

Note that the build on nodejs@8 will fail with lint errors until PR gulpjs/eslint-config-gulp#13 is merged.   That change is needed for newer versions of eslint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: required
 language: node_js
 node_js:
+  - node # latest stable node build
   - '6'
   - '5'
   - '4'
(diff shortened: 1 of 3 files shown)


BurtHarris:
@phated, if you really really want `npm test` to run lint, the fix is simple.   Change these CI scripts to use `mocha -A` rather than `npm test`.

BurtHarris:
Needs PR gulpjs/eslint-config-gulp#13

phated:
Sorry to be a naysayer but not having everything work consistently across node 0.10-latest makes this a hard "no" from me.

BurtHarris:
OK, it your call.   Good luck with that.

PR #277, opened 2017-08-31
Options refactor

erikkemperman:
Cleaned up (IMHO) options handling slightly, in particular prevent users accidentally doing something odd with `flag`. Also, added an `append` option (which I think might come in handy for things like build logs).

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,14 @@ Type: `Boolean`
 
 Default: `true` (always overwrite existing files)
 
+##### `options.append`
+
+Whether or not new data should be appended after existing file contents (if any).
+
+Type: `Boolean`
+
+Default: `false` (always replace existing contents, if any)
+
 ##### `options.sourcemaps`
 
 Enables sourcemap support on files passed through the stream.  Will write inline soucemaps if specified as `true`.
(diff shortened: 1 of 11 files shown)


phated:
@erikkemperman so this takes away the ability to use `r`, `r+`, `w+`, etc?

erikkemperman:
Well, yes. None of those make sense in `dest`, do they? The file descriptor is always closed after (exclusively) writing or appending to it. The `flags` option was undocumented, and I figured my change only made explicit (and with clearer semantics than `flags`, IMHO) the already implemented ability to handle `a` in addition to `w`.

But if you expect some of those other flags might ever be useful, that's fine obviously; but I'd say it ought to be documented then?

phated:
@erikkemperman you are right and I don't think the `flags` option was exposed until our options changes.  I was just making sure it takes those away because they are still being tested for the utils.  I **really**, **really** want to factor the utils into their own modules :frowning_face:

phated:
@erikkemperman can you rebase this on top of the changes from the symlink/junction refactor (in master now)?  Thanks.

erikkemperman:
@phated Done!

phated:
Thanks @erikkemperman - This is great!  I've merged with a rebase to change some of the commit messages.

phated:
On second thought, I might squash those all into 1

PR #278, opened 2017-08-31
Bump deps

erikkemperman:
@phated Sorry if this redundant, I know you were also visiting deps. Still, the value-or-function upgrade turned out to be slightly quirky so wanted to submit this just in case.

### DIFF
--- a/lib/file-operations.js
+++ b/lib/file-operations.js
@@ -73,12 +73,12 @@ function getModeDiff(fsMode, vinylMode) {
 
 function getTimesDiff(fsStat, vinylStat) {
 
-  var mtime = date(vinylStat.mtime);
+  var mtime = date(vinylStat.mtime) || null;
   if (!mtime) {
     return;
   }
 
-  var atime = date(vinylStat.atime);
+  var atime = date(vinylStat.atime) || null;
   if (+mtime === +fsStat.mtime &&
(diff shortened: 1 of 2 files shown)


lib/file-operations.js
```diff
@@ -73,12 +73,12 @@ function getModeDiff(fsMode, vinylMode) {
 
 function getTimesDiff(fsStat, vinylStat) {
 
-  var mtime = date(vinylStat.mtime);
+  var mtime = date(vinylStat.mtime) || null;
```
phated: Why did this happen?

erikkemperman: Because value-or-function now returns `undefined` for invalid input, and the expression below this line is expecting `null`: Oddly `+null` yields `0`, whereas `+undefined` yields `NaN`, for some reason.

phated: Maybe we should make them 0 then?

erikkemperman: Yeah, that would work. But now that I look at this again, the expressions with `+` seem slightly wrong here. Aren't we discounting `0` as a valid date (namely 1970-01-01 00:00:00)? Unlikely to ever come up, I guess, but still.

Anyway, I've amended to `0` for now.

phated:
Thanks @erikkemperman 😺

PR #279, opened 2017-09-30
Scaffold: LICENSE should not be executable

erikkemperman:
Noticed that LICENSE was executable, apparently already since it was first added.

Also, perhaps prior to release it would make sense to copy this one to all the repos? Or alternatively update the year-ranges for the other ones?


phated:
Nice catch.  We should create some tooling to fix these (and the other PRs you have sent) across repos.  Even if it's just a bash script or something.

PR #282, opened 2017-11-07
Fix bom for buffers

sielay:
Current implementation cause error as `strip-bom` doesn't support buffers anymore. Hope it will pass on Travis, as locally (OSX sierra) one test about time `getTimesDiff makes atime diff undefined if fs and vinyl atime are invalid:` is not passing on your branch when it passes on travis.

### DIFF
--- a/lib/src/getContents/bufferFile.js
+++ b/lib/src/getContents/bufferFile.js
@@ -2,6 +2,7 @@
 
 var fs = require('graceful-fs');
 var stripBom = require('strip-bom');
+var stripBomBuffer = require('strip-bom-buf');
 
 function bufferFile(file, opt, cb) {
   fs.readFile(file.path, function(err, data) {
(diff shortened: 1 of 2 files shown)


sielay:
Please ignore it for a while. `strip-bom-buf` expose es6 code and your test work in strict mode.

erikkemperman:
@sielay the master branch switched to another BOm stripping utility: https://github.com/gulpjs/vinyl-fs/commit/b1d700afdb9d073a89d38db80aa1f4299bdea52a

Also, I'm not entirely sure but I think the `nighmare-gulp-sourcemaps` branch is no longer used.

sielay:
@erikkemperman it exposes 2.4.4, when your master is 2.4.3. It's easy to get that branch "by mistake" ;)

phated:
2.4.4 is the current published version that locked gulp-sourcemaps to a certain version.  The way semver works makes this PR unnecessary and we've replaced the dependencies for 3.0

PR #283, opened 2017-11-10
Remove unused require

erikkemperman:
Little blight left over from the symlink melee :-)


phated:
Landed this change in another cleanup commit.  Thanks for pointing it out.

Issue #288, opened 2017-12-15
Issues with libraries using an outdated vinyl version 


asgoth:
I don't really know if it belongs here, but I'm getting errors when upgrading to vinyl-fs version 3.x. 

Several libraries (the ones are found are gulp-watch and vinyl-source-stream) are using an outdated vinyl version, which causes errors using **vfs.dest**.

See

https://github.com/floatdrop/gulp-watch/issues/295
https://github.com/hughsk/vinyl-source-stream/issues/23

For now I've solved it internally by wrapping the outdated vinyl object in a new one:

```
import through = require('through2');
import File = require('vinyl');

/**
 * gulp-watch/vinyl-source-stream uses an outdated vinyl, which doesn't have some functions which vinyl-fs expects.
 * Can be removed when gulp-watch updates it to vinyl 2.x.
 * @returns {NodeJS.ReadWriteStream}
 */
export function upgrade(): NodeJS.ReadWriteStream {
  return through.obj(function(file: any, encoding: string, cb: through.TransformCallback): void {
    const upgradedFile = new File(file);
    cb(null, upgradedFile);
  });
}

```

phated:
@asgoth yeah, vinyl-fs 3 is a major, breaking version.  Plugins will need to be updated to use a version of Vinyl that supports `isVinyl()`

phated:
@erikkemperman do you think we should marshall vinyls from <2.x to vinyl 2+ objects?  It might solve this but I'm not sure what issues might be caused by doing that.

erikkemperman:
I’m not sure, but it feels a bit like a patch in the wrong place? Curing the symptom but not the disease. Might delay plugin adoption of vinyl 2 even, hiding this issue from sight.

But we could try first nudging the plugin devs, maybe submit a PR here and there? I’d guess the changes needed are not that hard, and probably similar from one plugin to the next. 

So once we have a few over, we could then just point to examples of what to do. I might have some time to try, would gulp-watch be a good candidate?

On the other hand, if new plugin releases are not forthcoming, and this is spoiling people’s first taste of the new vinyl-fs — perhaps a quick stopgap is in order...

phated:
@erikkemperman I've sent PRs to gulp-watch and vinyl-source-stream but haven't heard back yet.  The biggest problem seems to be gulp-util consumers (which is soft-deprecated currently)

phated:
Side note: I think I'm going to be using `npm-deprecate` against gulp-util today.

[...]

PR #291, opened 2017-12-22
Fix: Marshall old Vinyl objects to new Version in dest()

phated:
@erikkemperman @asgoth @stevelacy @doowb What do y'all think of this as the solution to #288 and gulpjs/gulp/issues/2065

### DIFF
--- a/lib/dest/prepare.js
+++ b/lib/dest/prepare.js
@@ -16,6 +16,10 @@ function prepareWrite(folderResolver, optResolver) {
       return cb(new Error('Received a non-Vinyl object in `dest()`'));
     }
 
+    if (typeof file.isSymbolic !== 'function') {
+      file = new Vinyl(file);
+    }
+
     var outFolderPath = folderResolver.resolve('outFolder', file);
     if (!outFolderPath) {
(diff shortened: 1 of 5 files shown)


doowb:
@phated LGTM! I like that a new Vinyl instance is created only when an old Vinyl instance is detected.

erikkemperman:
Not sure if we call `isSymbolic()` there, but if so should this be done for `symlink` and even `src` as well? And perhaps add a comment that this is to deal with out-of-date plugins and hopefully could be removed at some point?

stevelacy:
I do also agree with @erikkemperman relating to also check/modify `symlink`. The rest looks good

phated:
@erikkemperman @stevelacy I've added the marshalling to `.symlink` but I don't think it's necessary because we don't check `.isSymbolic()` in that API - (`.src` actually always marshalls in the `wrap-file` stream).  I've also added erik's requested TODO.

erikkemperman:
@phated Looks good to me, thanks!

phated:
@erikkemperman @doowb @stevelacy Thanks for reviewing this!

Issue #295, opened 2018-01-17
dest() result stream does not end when writing a stream using Node 9


mgroenhoff:
When writing a stream (`file.isStream()` return `true`) using `vfs.dest()` the resulting stream will never end. I did some digging and it seems the code inside `file-operations.js` is out of date with when compared to Node 9.

The issue is on the folowing line:
https://github.com/gulpjs/vinyl-fs/blob/e739f6c62ef69f53fe9ca210cbc79ac548ec1bf2/lib/file-operations.js#L376-L377

When i compare this to Node 9:
https://github.com/nodejs/node/blob/8c00a809bc08d87776f74c84751607155f3df61f/lib/fs.js#L2462-L2467

It calls destroy instead of close (which btw is not even part of the public API). When i change it to `this.once('finish', this.destroy);` (like Node does) everything seems to works fine.

phated:
As I stated in the other thread:
> unfortunately, we don't support node 9 because the node team doesn't seem to care that they keep breaking userland.

The logic you are referencing was pulled out of an LTS version of node and has worked across all versions until 9.  We can't keep bending to node-core's whim when we also have to think about backwards compat.

PR #296, opened 2018-01-17
Do not rely on Node.js internals

mcollina:
This restores compatibility with Node 9 by lifting the code from Node 8.9.4 and copying and pasting it here. This should have been the approach from the very beginning, because vinyl-fs is creating its own WriteStream that has very little in common with the one from Node core.

Unfortunately, pulling functions from a prototype and attaching them to a non-related one is a very dangerous technique: as you can imagine a function attached to a prototype relies on the state that is created by the relative constructor. These are implementation details, and they can surely change across major versions.

As a side note, this was likely broken by https://github.com/nodejs/node/pull/15407 and https://github.com/nodejs/node/pull/18002.

Fixes https://github.com/gulpjs/vinyl-fs/issues/295

---

The actual logic can probably be simplified, because vinyl-fs `WriteStream` implementation is different than the one in Node core.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: required
 language: node_js
 node_js:
+  - '9'
   - '8'
   - '6'
   - '5'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,7 @@ environment:
     - nodejs_version: "5"
     - nodejs_version: "6"
     - nodejs_version: "8"
+    - nodejs_version: "9"
 
 install:
   - ps: Install-Product node $env:nodejs_version
--- a/lib/file-operations.js
+++ b/lib/file-operations.js
@@ -398,10 +398,47 @@ WriteStream.prototype.open = function() {
 
 // Use our `end` method since it is patched for flush
 WriteStream.prototype.destroySoon = WriteStream.prototype.end;
-// Use node's `fs.WriteStream` methods
-WriteStream.prototype._destroy = fs.WriteStream.prototype._destroy;
-WriteStream.prototype.destroy = fs.WriteStream.prototype.destroy;
-WriteStream.prototype.close = fs.WriteStream.prototype.close;
+
+WriteStream.prototype._destroy = function(err, cb) {
+  this.close(function(err2) {
+    cb(err || err2);
+  });
+};
+
+WriteStream.prototype.close = function(cb) {
+  var that = this;
+
+  if (cb) {
+    this.once('close', cb);
+  }
+
+  if (this.closed || typeof this.fd !== 'number') {
+    if (typeof this.fd !== 'number') {
+      this.once('open', closeOnOpen);
+      return;
+    }
+
+    return process.nextTick(function() {
+      that.emit('close');
+    });
+  }
+
+  this.closed = true;
+
+  fs.close(this.fd, function(er) {
+    if (er) {
+      that.emit('error', er);
+    } else {
+      that.emit('close');
+    }
+  });
+
+  this.fd = null;
+};
+
+function closeOnOpen() {
+  this.close();
+}
 
 function worker(data, encoding, callback) {
   var self = this;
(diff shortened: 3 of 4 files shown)


phated:
@mcollina thanks!  It looks like these changes devastated Windows.   Any ideas?

mcollina:
@phated I've just checked on a Windows box, and it's failing even on master with Node 8 and 6. It's not related (but should probably be fixed anyway in another PR).

phated:
@mcollina not sure about that https://ci.appveyor.com/project/gulpjs/vinyl-fs/build/735 - we've had successful builds for quite some time.

mcollina:
@phated these are what I am getting (more or less all of the same, ENOENT on files):

```
  24) symlink stream (windows) options can disable junctions for a directory (as a function):
     Uncaught Error: ENOENT: no such file or directory, stat 'C:\Users\Matteo\Repositories\vinyl-fs\test\out-fixtures\foo'
      at Error (native)
      at Object.fs.statSync (fs.js:1009:11)
      at Object.statSync (node_modules\graceful-fs\polyfills.js:297:22)
      at assert (test\symlink.js:395:22)
      at ConcatStream.<anonymous> (node_modules\concat-stream\index.js:36:43)
      at finishMaybe (node_modules\readable-stream\lib\_stream_writable.js:607:14)
      at endWritable (node_modules\readable-stream\lib\_stream_writable.js:615:3)
      at ConcatStream.Writable.end (node_modules\readable-stream\lib\_stream_writable.js:571:41)
      at ConcatStream.Writable._destroy (node_modules\readable-stream\lib\_stream_writable.js:662:8)
      at ConcatStream.destroy (node_modules\readable-stream\lib\internal\streams\destroy.js:36:8)
      at node_modules\mississippi\node_modules\pump\index.js:43:45
      at call (node_modules\mississippi\node_modules\pump\index.js:50:3)
      at Array.forEach (native)
      at node_modules\mississippi\node_modules\pump\index.js:70:25
      at f (node_modules\once\once.js:25:25)
      at Pumpify.<anonymous> (node_modules\mississippi\node_modules\pump\index.js:29:21)
      at Pumpify.f (node_modules\once\once.js:25:25)
      at Pumpify.onerror (node_modules\end-of-stream\index.js:43:12)
      at Pumpify.Duplexify._destroy (node_modules\duplexify\index.js:195:15)
      at node_modules\duplexify\index.js:179:10
      at _combinedTickCallback (internal/process/next_tick.js:73:7)
      at process._tickCallback (internal/process/next_tick.js:104:9)
```

(this is node 6 on master)

Maybe a dependency changed and broke things?

mcollina:
Can you trigger a build on master?

phated:
@mcollina done and still passing on Windows: https://ci.appveyor.com/project/gulpjs/vinyl-fs/build/738

mcollina:
@phated the test suite on master fails on Windows Home Edition. I guess it's due to the permissions required to create symlinks (I stumbled upon that in the past, and it seems the case).

I ended up removing FlushWriteStream and using Writable directly (since readable-stream 2.3.0 it offer `_final`). The alternative is to copy the methods from Node 6 instead, those should work. I went for Writable, because it removes a dependency.
Let me know if you prefer the Node 6 approach.

phated:
> I ended up removing FlushWriteStream and using Writable directly (since readable-stream 2.3.0 it offer _final)

Oh wow, so `_final` is always called before the defined `.close` method is called (since that closes the file descriptor)?  That's a great additional API!

mcollina:
> Oh wow, so _final is always called before the defined .close method is called (since that closes the file descriptor)? That's a great additional API!

Yes! `_final` is called before the streams emits `'finish'`, it was such a useful feature that we added it in Node 8 (readable-stream@2.3.x).

phated:
This was rebased and merged as 167d8ffdcbbade502ba3175d3e224fa83ea830e0 and 62d196961f5814999e334abdf4b919befe32b632 - publishing as 3.0.2 once CI completes.

Thanks again!

PR #304, opened 2018-10-14
Do not use the deprecated Buffer() constructor

lpinca:
Refs: https://github.com/nodejs/citgm/issues/605
Refs: https://nodejs.org/api/deprecations.html#deprecations_dep0005_buffer_constructor

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "jscs-preset-gulp": "^1.0.0",
     "mississippi": "^1.2.0",
     "mocha": "^3.5.0",
-    "rimraf": "^2.6.1"
+    "rimraf": "^2.6.1",
+    "safer-buffer": "^2.1.2"
   },
   "keywords": [
     "gulp",
--- a/test/dest-modes.js
+++ b/test/dest-modes.js
@@ -4,6 +4,7 @@ var fs = require('graceful-fs');
 var File = require('vinyl');
 var expect = require('expect');
 var miss = require('mississippi');
+var saferBuffer = require('safer-buffer');
 
 var vfs = require('../');
 
@@ -18,6 +19,8 @@ var testConstants = require('./utils/test-constants');
 var from = miss.from;
 var pipe = miss.pipe;
 var concat = miss.concat;
+
+var Buffer = saferBuffer.Buffer;
 
 var inputBase = testConstants.inputBase;
 var outputBase = testConstants.outputBase;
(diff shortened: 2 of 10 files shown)


lpinca:
@phated are you interested in this?

phated:
@lpinca I'm going to be tweaking stuff in the next week or so.  It looks like there are some file conflicts right now.

lpinca:
> It looks like there are some file conflicts right now.

Yes, I asked because exactly for that reason. There is no point in resolving conflicts if the patch is not wanted.

BurtHarris:
Seems like this is languishing due to a very narrow test not failing _in the expected way!_   An assert IS generated is appveyor test environment (not travis-ci) running Node 0.12, perhaps the test is too specific as to what is expected in a negative test.  

What is the point of supporting Node.js version all the way back to version 0.   I somehow doubt anyone is seriously interested in building systems that depend on that platform, particularly on Windows. 

Stack for that failure suggests to me the bug is in the test.   It says:

---
  1) closeFd calls the callback with close error if no error to propagate:
     Uncaught Error: Expected undefined to exist
      at assert (node_modules\expect\lib\assert.js:29:9)
      at Expectation.toExist (node_modules\expect\lib\Expectation.js:52:28)
      at test\file-operations.js:733:19
      at onClosed (lib\file-operations.js:26:5)
      at node_modules\graceful-fs\graceful-fs.js:45:10
      at FSReqWrap.oncomplete (fs.js:95:15)

phated:
@BurtHarris it's not been looked at because I started a new job that's taking most of my time and traveling which is taking the rest. And I'm the only one "actively" working on this stuff.

Thanks for diving in further, but I believe that test was a regression test so we should probably figure out why it is breaking (though it probably isn't happening from this PR???).

phated:
@sttk would you have some time to debug that test?

sttk:
@phated I could reproduce this error on appveyor but not on my Windows environment. This error may be appveyor specific. I'll dig into detail about this.

sttk:
This problem on v0.12 is the known issue as per [https://github.com/nodejs/node-v0.x-archive/issues/4574](https://github.com/nodejs/node-v0.x-archive/issues/4574) or [https://github.com/nodejs/node/issues/3718](https://github.com/nodejs/node/issues/3718).
And this problem happens on node.js v0.11, v0.12 and io.js.
So [the target test case](https://github.com/gulpjs/vinyl-fs/pull/304/files#diff-d47dcce7a4fb075fa8beab4006af26faL724) should be skipped if node version > 0.10 and < 4.

phated:
Since we are dropping node <10.13.0 support, we will be able to use the buffer constructor directly.

PR #314, opened 2019-12-25
Update: Remove meaningless parameter

skyline-123:
Update: Remove meaningless parameter


phated:
This was intentional to keep everything uniform.

PR #322, opened 2021-01-15
Breaking: Upgrade scaffold, dropping node <10 support

coreyfarrell:


coreyfarrell:
I'm confused, it failed on Windows + Node.js 10 only.  I don't have easy access to any machine running Windows, not sure if anyone can help troubleshoot this.

phated:
I honestly wouldn't be surprised if the node team did something weird with the symlink behavior between 8 and 10 and 12 on Windows.

coreyfarrell:
> I honestly wouldn't be surprised if the node team did something weird with the symlink behavior between 8 and 10 and 12 on Windows.

I just opened #323 to test on Github Actions without any of the other scaffold changes, looks like this test failure is not due to changes in this PR.

WikiRik:
Hi all, any chance we can get an update on this? I have a Dependabot alert on `glob-parent` I would like to get rid of. Thanks to your great work on `glob-stream` it's just this PR that's needed.

The logs have expired for the failing test on Windows, but if they are run again I might be able to take a look at it

Thanks again for your work on these great libraries!

phated:
@WikiRik  This is not the only thing that needs to be done. Corey was leading the charge on updating the packages to streamx but life stuff came up (as with many people during the pandemic) so it's on hold. None of us do this as a job.

WikiRik:
I understand, no worries. Thanks again for your hard work!

phated:
Replacing this with #333

PR #329, opened 2022-05-17
Dependency upgrades

kinetifex:
Contribution package upgrades, related to this discussion: https://github.com/gulpjs/gulp/discussions/2590

Looks like the work to update `devDependencies` was being tackled in https://github.com/gulpjs/vinyl-fs/pull/322.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "remove-bom-buffer": "^3.0.0",
     "remove-bom-stream": "^1.2.0",
     "resolve-options": "^1.1.0",
-    "through2": "^2.0.0",
+    "through2": "^4.0.2",
     "to-through": "^2.0.0",
     "value-or-function": "^4.0.0",
     "vinyl": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "fs-mkdirp-stream": "^1.0.0",
     "glob-stream": "^7.0.0",
     "graceful-fs": "^4.0.0",
-    "iconv-lite": "^0.4.24",
+    "iconv-lite": "^0.6.3",
     "is-valid-glob": "^1.0.0",
     "lazystream": "^1.0.0",
     "lead": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "lazystream": "^1.0.0",
     "lead": "^3.0.0",
     "object.assign": "^4.0.4",
-    "pumpify": "^1.3.5",
+    "pumpify": "^2.0.1",
     "readable-stream": "^3.6.0",
     "remove-bom-buffer": "^3.0.0",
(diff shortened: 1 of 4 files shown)


phated:
Thanks, but dependency upgrades get handled by our updater tool. They are strictly semver managed.

package.json
```diff
@@ -27,21 +27,21 @@
...
     "lazystream": "^1.0.0",
-    "lead": "^1.0.0",
+    "lead": "^3.0.0",
     "object.assign": "^4.0.4",
-    "pumpify": "^1.3.5",
-    "readable-stream": "^2.3.3",
+    "pumpify": "^2.0.1",
+    "readable-stream": "^3.6.0",
     "remove-bom-buffer": "^3.0.0",
     "remove-bom-stream": "^1.2.0",
```
kinetifex, line 40: So, everything upgraded smoothing expect I have run into some issues with [remove-bom-stream v2](https://github.com/gulpjs/remove-bom-stream) which was published not too long ago.

I haven't delved too deep into it yet, but wanted to reach out if there's some early guidance for tackling this @phated? Is kinda the goal to move from `through2` to `streamx` across-the-board (per https://github.com/gulpjs/remove-bom-stream/issues/4)?

phated: Yes, almost the entirety of the gulp 5 holdup is finding someone to do the streamx updates across all projects.

test/dest.js
```diff
@@ -945,6 +945,7 @@ describe('.dest()', function() {
     pipe([
       from.obj([file]),
       destStream,
+      slowCount(1), // Verify and avoid race condition w/ assert check
```
kinetifex, line 948: I seemed to start having failing assets after upgrading `readable-stream` to v3, including the packages that had v2 as a sub dependency. It feels like a race condition with these tests that I left code comments on. Does this test fix make sense?

test/integration.js
```diff
@@ -55,10 +52,15 @@ describe('integrations', function() {
     }
 
     pipe([
-      vfs.src(inputGlob, { buffer: false }),
+      vfs.src(inputGlob, { highWaterMark: expectedCount }),
```
kinetifex, line 55: I could not find `buffer` in use. `highWaterMark` is not a documented option, but is used in code and other tests, so I went with it.

package.json
kinetifex: I see - would you be open to an intermediate release for some of these upgrades to `vinyl-fs` for now? Maybe leave the streamx work out for now but catch up on some of these other lagging dependencies. I left some other comment questions, but so far it doesn't seem like anything would constitute a major breaking change.

phated: Every major causes a major. All of the ones in the gulpjs org dropped node support. Everything leads to gulp 5 with no options for "intermediate"

PR #333, opened 2023-03-19
feat!: Switch stream implementation to streamx

sttk:
This PR inherits #322 and updates this repository more for the following points:

1. Upgrades dependencies and drops node <10.13 support
    - `rimraf`'s version is 3, not latest, because the latest version 4 supports only node>=14.
    - `eslint`, `expect`, and `mocha` upgrades to the last versions which supports node v10.
    - Other dependencies upgrades latest versions.

2. Drops `through2` and makes use of `streamx`

3. Drops `mississippi` and `readable-stream`

4. Drops `remove-bom-buffer`, but `is-utf8` is used.

5. Skips 2 test cases about directory symlink on Node.js v10 and Windows

    Because `EPERM` error occurs only on this environment. 

In addition, I tryed to drop `pumpify` by using `streamx`, etc. in `.dest()` and `.symlink()` but cloudn't do them.

NOTE: This package depends on [`glob-stream`](https://github.com/gulpjs/glob-stream) which will be bumped up.


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2013-2017 Blaine Bublitz <blaine.bublitz@gmail.com>, Eric Schoffstall <yo@contra.io> and other contributors
+Copyright (c) 2013-2020, 2023 Blaine Bublitz <blaine.bublitz@gmail.com>, Eric Schoffstall <yo@contra.io>.
```
phated, line 3:
```suggestion
Copyright (c) 2013-2020, 2023 Blaine Bublitz <blaine.bublitz@gmail.com> and Eric Schoffstall <yo@contra.io>.
```

lib/codecs.js
```diff
@@ -3,7 +3,7 @@
 /* eslint-disable new-cap */
 
 var iconv = require('iconv-lite');
-var through = require('through2');
+var Transform = require('stream').Transform;
```
phated, line 6: We'll want this to be streamx

lib/src/index.js
```diff
@@ -20,16 +20,14 @@ function src(glob, opt) {
     throw new Error('Invalid glob argument: ' + glob);
   }
 
-  var streams = [
+  var outputStream = pipeline(
```
phated, line 23: Does pipeline actually work for this? I didn't think pipeline would emit any files from the end

lib/src/read-contents/read-buffer.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 var fs = require('graceful-fs');
-var removeBomBuffer = require('remove-bom-buffer');
+var isUtf8 = require('is-utf8');
```
phated, line 4: Did we have a way around using this in removeBomStream?

package.json
```diff
@@ -19,43 +19,47 @@
...
     "pretest": "npm run lint",
-    "test": "mocha --async-only",
-    "cover": "istanbul cover _mocha --report lcovonly",
-    "coveralls": "npm run cover && istanbul-coveralls"
+    "test": "nyc mocha --async-only"
   },
   "dependencies": {
-    "fs-mkdirp-stream": "^1.0.0",
+    "fs-mkdirp-stream": "^2.0.1",
     "glob-stream": "^6.1.0",
```
phated, line 28:
```suggestion
    "glob-stream": "^8.0.0",
```

package.json
```diff
@@ -19,43 +19,47 @@
...
-    "readable-stream": "^2.3.3",
-    "remove-bom-buffer": "^3.0.0",
-    "remove-bom-stream": "^1.2.0",
-    "resolve-options": "^1.1.0",
-    "through2": "^2.0.0",
-    "to-through": "^2.0.0",
-    "value-or-function": "^3.0.0",
-    "vinyl": "^2.0.0",
-    "vinyl-sourcemap": "^1.1.0"
+    "lazystream": "^1.0.1",
```
phated, line 33: We want to drop this since it's not streamx and the maintainer is absent

package.json
```diff
@@ -19,43 +19,47 @@
...
+    "lead": "^4.0.0",
+    "pumpify": "^2.0.1",
```
phated, line 35: Want to drop pumpify, I'll try to get `compose` into streamx

package.json
```diff
@@ -19,43 +19,47 @@
...
-    "eslint-config-gulp": "^2.0.0",
-    "expect": "^1.19.0",
-    "istanbul": "^0.4.3",
-    "istanbul-coveralls": "^1.0.3",
-    "jscs": "^2.4.0",
-    "jscs-preset-gulp": "^1.0.0",
-    "mississippi": "^1.2.0",
-    "mocha": "^3.5.0",
-    "rimraf": "^2.6.1"
+    "concat-stream": "^2.0.0",
```
phated, line 45: Want to drop this

test/codecs.js
```diff
@@ -2,11 +2,14 @@
...
+var concat = require('concat-stream');
 
-var from = miss.from;
-var pipe = miss.pipe;
-var concat = miss.concat;
+// TODO: This `from` should be replaced to `node:stream.Readable.from`
+// if this package supports only >= v10.17.0
+var from = require('streamx').Readable.from;
+
+var pipeline = stream.pipeline;
```
phated, line 12: Tests need to be rewritten to check multiple types of streams

test/src-symlinks.js
```diff
@@ -55,13 +59,19 @@ describe('.src() with symlinks', function() {
...
       concat(assert),
     ], done);
   });
 
   it('resolves directory symlinks correctly', function(done) {
+    // Error: EPERM: operation not permitted
+    if (os.platform() === 'win32' && nodeVersion.major === 10) {
+      this.skip();
+      return;
+    }
```
phated, line 73: I want to look into this to understand it

test/utils/test-streams.js
phated, line 1: These utils will need to be based on the streams being tested, like the other streaming modules that switched to streamx

phated (changes requested):
Thanks for updating this. I've made some comments, but they are mostly for me to remember what needs to be done to finish up this PR

lib/codecs.js
sttk: @phated 
This `Tansform` operates text streams in text mode. while `streamx` operates binary streams in object mode.

lib/src/read-contents/read-buffer.js
sttk: `removeBomStream` was changed to require encoding by `removeBomStream#8`. 
Before the change, `removeBomStream` was using `removeBomBuffer`.

lib/codecs.js
phated: @sttk that's incorrect, streamx switches between object and binary automatically. you can receive/send either

package.json
phated: Waiting on https://github.com/mafintosh/streamx/pull/73

package.json
```diff
@@ -19,43 +19,48 @@
...
-    "coveralls": "npm run cover && istanbul-coveralls"
+    "test": "nyc mocha --async-only"
   },
   "dependencies": {
-    "fs-mkdirp-stream": "^1.0.0",
-    "glob-stream": "^6.1.0",
-    "graceful-fs": "^4.0.0",
-    "iconv-lite": "^0.4.24",
+    "fs-mkdirp-stream": "^2.0.1",
+    "glob-stream": "^8.0.0",
```
ssiltanen, line 28: This fixes a high severity vulnerability with glob-parent https://www.mend.io/vulnerability-database/CVE-2020-28469
If this PR has issues and cannot be merged, it would be nice to get this security patch sooner.

test/file-operations.js
```diff
@@ -76,24 +80,27 @@ describe('isOwner', function() {
...
 
     // Windows :(
-    if (process.getuid === noop) {
-      delete process.getuid;
+    if (typeof process.getuid !== 'function') {
+      process.getuid = noop;
     }
 
+    sinon.restore();
+
```
phated, line 103: I don't understand this. Flagging to look at myself.

test/src.js
```diff
@@ -128,35 +131,35 @@ describe('.src()', function() {
...
   });
 
   it('removes BOM from utf8-encoded files by default (buffer)', function(done) {
-    var expectedContent = new Buffer(bomContents.replace('X', '8'));
+    var expectedContent = Buffer.from(bomContents.replace('X', '8'));
 
     function assert(files) {
       expect(files.length).toEqual(1);
-      expect(files[0].contents).toMatch(expectedContent);
+      expect(files[0].contents.toString()).toEqual(expectedContent.toString());
```
phated, line 145: Need to remove these toString calls. Noting to remove when I refactor this file.

test/src.js
```diff
@@ -211,36 +217,39 @@ describe('.src()', function() {
...
     }
 
-    pipe([
+    pipeline([
       vfs.src(bomInputPath, { removeBOM: false, buffer: false }),
-      through.obj(compareContents),
+      new stream.Transform({
+        objectMode: true,
+        transform: compareContents,
+      }),
```
phated, line 225: Probably a utility

test/src.js
```diff
@@ -709,10 +745,10 @@ describe('.src()', function() {
       expect(files.length).toEqual(1);
       expect(files[0].path).toEqual(inputPath);
       expect(files[0].isNull()).toEqual(true);
-      expect(files[0].contents).toNotExist();
+      expect(files[0]).toHaveProperty('contents', null);
```
phated, line 748:
```suggestion
      expect(files[0].contents).toEqual(null);
```
I think?

test/symlink.js
```diff
@@ -704,62 +706,78 @@ describe('symlink stream', function() {
...
-    symlinkStream.on('end', done);
+    var ended = false;
+    function end() {
+      ended = true;
+    }
+
+    symlinkStream.on('end', function() {
+      expect(ended).toBeTruthy();
+      done();
+    });
```
phated, line 717: I think this is the wrong order, we want to se the flag in the event and handle the assertion in the callback to pipeline

test/symlink.js
```diff
@@ -704,62 +706,78 @@ describe('symlink stream', function() {
...
-    symlinkStream.on('finish', done);
+    var finished = false;
+    symlinkStream.on('finish', function() {
+      finished = true;
+    });
+
+    function finish() {
+      expect(finished).toBeTruthy();
+      done();
+    }
```
phated, line 742: same as above

test/symlink.js
```diff
@@ -824,17 +845,25 @@ describe('symlink stream', function() {
...
         readables++;
       }
+    }).on('finish', function() {
+      finished = true;
+      expect(readables).toEqual(0);  // `readable` was not emitted
+    }).on('end', function() {
+      ended = true;
+      expect(readables).toEqual(1);  // `readable` was emitted
+      done();
     });
```
phated, line 856: these are broken

test/file-operations.js
sttk: It's my mistake. I've corrected this part and added an explanation comment.

test/src.js
sttk: I removed `.toString()` here and others in `test/src.js`.
But some places were not removed because data types of them are object and they caused errors by removing.

test/symlink.js
sttk: I embeded `console.log` into `end` and the event handler of `symlinkStream:end`, and the order of calls were `end() -> symlinkStream:end`.

If this order looks strange, should I modify this part as follows?
```
  symlinkStream.on('end', done);
  function end() {}
```

test/symlink.js
sttk: `finish` event is called before `pipeline`'s last callback.
(`symlinkStream:finish -> finish()`)

test/src.js
phated: The tests that were failing seem to be due to an incorrect `concat` implementation. I refactored the stream utils a bit.

lib/src/read-contents/read-stream.js
```diff
@@ -16,24 +16,25 @@ function streamFile(file, optResolver, onRead) {
 
   var filePath = file.path;
 
-  file.contents = new lazystream.Readable(function() {
+  file.contents = new lazystream.Readable(function () {
```
phated, line 19: Need to get rid of lazystream (will do in follow up PR).

phated: #336

package.json
phated: #336

test/file-operations.js
phated: This looks good to me now.

test/src-symlinks.js
phated: This seems to be a bug with node 10 but we want to support Node 10 on Windows so I've written a workaround. I left a TODO to remove the recursive readlink when we drop Node 10 and I'll open a follow up issue.

phated (approved):
Tests should be green now and I'm happy enough with the diff. Thanks for getting this started @coreyfarrell and thanks for updating @sttk 🎉  This truly was a team effort.

Issue #351, opened 2024-04-25
Vinyl-fs src corrupts image files


vkorytska:
While the issue is reproduced for Gulp.src, not for Vinyl.src, but this is relevant because Gulp.src uses Vinyl.src.

### What were you expecting to happen?

The task should copy images without issue.

### What actually happened?

The task corrupted the images. Similar corruption does not occur for text-based files. '{ removeBOM: false, buffer: false, encoding: false }' arguments do not fix the issue.

### Please give us a sample of your gulpfile

I created a (public) repository to show the issue with src and images: https://github.com/vkorytska/gulp-src-issue.

### Terminal output / screenshots

Added examples of corrupted images to the repository.

### Please provide the following information:
* OS & version [e.g. MacOS Catalina 10.15.4]: Windows 10 22H2
* node version (run `node -v`): v21.7.3
* npm version (run `npm -v`): 10.5.2
* gulp version (run `gulp -v`): 5.0.0

### Additional information

This issue is probably similar to https://github.com/aws/aws-sdk-js/issues/1628. In the repository added a quick fix of replacing fs.createReadStream with fs.readFile and the images were copied properly.

I am not sure why specifically image files are corrupted by createReadStream, and not text files.

tommyZZM:
👍🏽 yes same issue occur to me

pavelhoral:
I was able to recreate the issue when `gulp.src` had the same options as the in the original post (`{ removeBOM: false, buffer: false, encoding: false }`). However the issue can not be reproduced with the following options `{ encoding: false }`.

Having `removeBOM` does not make sense for `encoding: false`. But the issue is present when `buffer: false` (so some sort of bug is present there).

ImKubass:
I am having same issue

PR #352, opened 2024-05-30
Binary stream corruption fix

PRR24:
Addresses https://github.com/gulpjs/gulp/issues/2803

### DIFF
--- a/lib/dest/write-contents/write-stream.js
+++ b/lib/dest/write-contents/write-stream.js
@@ -31,7 +31,7 @@ function writeStream(file, optResolver, onWritten) {
   // TODO: should this use a clone?
   var streams = [file.contents];
 
-  if (encoding && encoding.enc !== DEFAULT_ENCODING) {
+  if (encoding && codec.enc !== DEFAULT_ENCODING) {
     streams.push(getCodec(DEFAULT_ENCODING).decodeStream());
     streams.push(codec.encodeStream());
   }


lubber-de (approved):

Daniel-Sh-uq:
Seems related to https://github.com/gulpjs/vinyl-fs/issues/351

phated (changes requested):
Needs regression tests still

nicolashenry:
I think that the test should be the same as [this test](https://github.com/gulpjs/vinyl-fs/blob/ad025afed42f70a064801180185b500235d5272a/test/dest.js#L712) by just removing "encoding: false".

erikkemperman:
This fix seems "obviously" correct -- somehow this typo sneaked through the whole vetting or there was some merging weirdness. But I don't yet understand why this branch of the code was never reached by the tests. It's been a long time since I looked at this and there's some things I don't recognize (such as a suite of stream backends, is that right?)
It would be good to have a bunch of failing tests in the old (current) situation which this PR then fixes.

PRR24:
There are currently no e2e binary stream tests. Here is the example of failing test, please add to the most suitable test file.
```js
  it('should not destroy binary file', function (done) {
    var ranBomInputPath = path.join(testConstants.inputBase, './ranbom.bin');
    var ranBomOutputPath = path.join(testConstants.outputBase, './ranbom.bin');

    function assert(files) {
      var srcResult = fs.readFileSync(ranBomInputPath, null);
      var destResult = fs.readFileSync(ranBomOutputPath, null);
      expect(srcResult).toEqual(destResult);
    }

    pipeline(
      [
        vfs.src(ranBomInputPath, { encoding: false, buffer: false }),
        vfs.dest(testConstants.outputBase),
        concatArray(assert),
      ],
      done
    );
  });

woody-li:
Gentle ping, any progress on this?

PR #358, opened 2025-05-31
fix: Avoid Node.js deprecation warning for `fs.Stats`

phated:
The Node.js core team is cruel and moves things to their `internal/` package and then marks them as deprecation. We can work around this using some crazy JS techniques (seen below).

When logging the new `fs.Stats` created by this technique, they look the same as `new fs.Stats()` and no deprecation warning is shown.


yocontra (approved):

robertknight:
While reviewing a dependency update for Gulp I looked into the context of the `fs.Stats` change. It is explained in https://github.com/nodejs/node/pull/51681#issuecomment-1963048465, found via https://github.com/nodejs/node/pull/51879. This comment highlights that a possible future change may break the code here again:

> If we manage to remove it from the public API we can just do Stats extends StatsBase here (which is a breaking change by disallowing construct calls without new/Function.prototype.call

Razunter:
This change broke Deno compatibility:
````
error: Uncaught TypeError: Class constructor Stats cannot be invoked without 'new'
    at newStats (... /node_modules/.deno/vinyl-fs@4.0.2/node_modules/vinyl-fs/lib/dest/prepare.js:13:34)
```

phated:
@Razunter 

1. Gulp doesn't support Deno. You can see in our CI matrix that we only support Node. 
2. Seems like a broken compatibility layer in Deno. You should report it to them.

